PR #38, opened 2026-08-24
CANVAS CENSUS — 170/170 minted, 168 recognisable, 162 sets with recorded round-trip drift rows, 22 walls (+47 projection cells)

tpitre:
Step 2 of the approved finish plan: the CANVAS HALF of the census for every committed contract. `census:check --phase full` is GREEN — 170/170 sets minted on the Scratch Project (`byMp6lt0Ij9b2QbkDGFwBh`), 805/805 sampled-variant canvas shots beside their code renders, a per-set live dump → propose → diff record, and a scored verdict for every row. The fast lane is flipped from `--phase code` to `--phase full`.

## Per-library

| library | sets | canvas shots | recognisable | walls (non-projection) |
|---|---|---|---|---|
| first-party | 54 | 138 | 54/54 | — (slot-only containers render empty in BOTH halves, noted per row) |
| altitude | 8 | 39 | 8/8 | tertiary border code-absent; avatar code render blank (0 ink px) |
| antd | 12 | 60 | 12/12 | exam sets REUSED — re-running the re-emitted scripts reported all-SKIP (specHash unchanged); 3 code-render gaps named (input blank, progress/tag partial) |
| astryx | 13 | 73 | 13/13 | toast surface white-on-white |
| carbon | 10 | 64 | 10/10 | warning glyph fill lost on canvas; tertiary border; ring export |
| fluent | 11 | 92 | 10/11 | **fluent.input NOT recognisable** — outline/underline appearances render 0 ink on canvas |
| mui | 31 | 139 | 31/31 | text-field label order disagrees |
| polaris | 12 | 67 | 12/12 | text/text-field are curated projections — 47 sampled cells named `CANVAS-PROJECTION:<variant>` (drawn cells verified by set export) |
| shadcn | 11 | 45 | 10/11 | **shadcn.input NOT recognisable** — leaf input with no text prop hugs to 288×8 and draws no stroke; border-color unspecified at base on 4 sets |
| tailwind | 8 | 41 | 8/8 | Active/Focus ring carried as effect data but unpainted in COMPONENT export (`FC-EXPORT-RING-UNPAINTED`) |

## Round-trip (dump → propose → diff vs the committed contract)

[...]

PR #39, opened 2026-08-25
The rejected sets, fixed at their causes — 6 conformance cases red→green, 29 sibling sets moved

tpitre:
Seven census sets the owner rejected by screenshot, each diagnosed to its true cause and fixed there — never at the symptom. Six new conformance cases red→green; the fixes move sibling sets across four libraries.

## Per-set cause chains

**1. fluent.card — "Preview" + header/description glued on one row.** Two capture losses, both in the shared floor pipeline. (a) `flex-direction` varied on the `orientation` axis and `enrichLayout` was uniform-or-nothing (`extract/computed/fuse.ts` "layout-not-uniform"), so the DEFAULT column direction was dropped and both surfaces drew the horizontal default. Fixed: single-axis factoring — base combo carries `layout.direction`, deviating values ride `layoutByProp` (the schema's own v7 spelling). (b) CardHeader is a real CSS grid whose resolved column list is `0px 404px` (the empty media column); the authored-zero fence (`grid-track-zero`, P2b) poisoned the computed-evidence plane and the whole grid promotion abandoned to the horizontal flex default. Fixed: computed 0px tracks collapse by name (`grid-zero-track-collapsed`) with placement re-indexing. Case: `flex-direction-by-axis` (WRONG-NAME→PASS) + `grid-zero-track-collapse` (WRONG-NAME→PASS).

**2. fluent.dialog — title/X/body/actions on one horizontal row.** DialogBody is `display:grid`, rows `28px 28px 32px`, columns `267px 267px 0px`. Three chained refusals: the zero-track fence (same as card), Chromium's span-1 serialization `grid-row: 1 / 1` refused as "end line must follow start line" (CSS §8.3.1 says an end line equal to the start line is span 1), and — once the grid finally carried — the emitter's grid placement throwing `Node is not a grid child` because a margin-carrying grid item is built inside its "(margin box)" wrapper and the placement setter targeted the inner node. All three fixed at cause (anatomy.ts ×2, emit-figma-script.ts `applyGridChildren` places the outermost wrapper). The dialog now stacks: title row (title + close action), body, actions.

**3. astryx.card — 13 hug-pill variants at 114×52.** The `width` token (114.203px) was the harness sample text's hug width minted as a design token: the block-root width door admitted any block root that measured neither the stage box nor the viewport, with a receipt that said "the number is the library's". It was not — a fresh re-fuse of the SAME library version measured 119.016px (a font swap moved the text metrics). Fixed: the shrink-to-fit branch is now a REFUSAL (`block-root-width-refused (shrink-to-fit)`); the canvas hugs, and `blockStage: true` is the named recapture remedy. Case: `block-root-shrink-to-fit` (UNDECLARED-CARRY→PASS). This also unblocked astryx's promote (see below).

**4. shadcn.checkbox — checked glyph a filled blob overflowing the 16px box.** Two halves of one cause. The floor's SVG reconstruction folded the lucide path's `fill: none` into the svg tag's inherited none and emitted NO fill attribute — but SVG's initial fill is BLACK, so the open check path filled solid the moment the asset rendered standalone; then `iconSvg` injected the paint as a root `fill`, drawing the blob in the check's ink. Fixed in `reconstructSvg` (fill="none" preserved by name — receipt `svg-stroke-glyph-fill-none`) and in `iconSvg` (a stroke-drawn glyph with no fill gets `fill="none"`, never the paint). Both shadcn assets re-derived from committed truth. Case: `svg-stroke-glyph-fill-none` (WRONG-NAME→PASS).

**5. shadcn.select — bare text+caret, no box.** The trigger's base border COLOUR was equal to the bare control element's, because shadcn's global `* { border-color: var(--border) }` styles the control too (and --input == --border in the theme) — the delta-from-control door dropped it as "not a fact of this component". The CSS surface still painted (border-color's initial value is currentColor); the canvas drew no stroke. Fixed: `resetSuppliedBorderColor` door (the exact twin of the existing border-STYLE reset door), receipted by name. Siblings: shadcn.alert and shadcn.input carried the same hole; both now carry base border colours. Case: `border-color-control-equal` (SILENT-LOSS→PASS).

**6. polaris.checkbox — "~3× scale, wrong tint": the premise did NOT survive re-measurement.** Both receipts render an 18 CSS-px box, 13px label, 28px row — Polaris truth exactly; the checked tint is #303030 on both surfaces. The two REAL defects found instead: (a) the check icon's `top/left` insets are a clean function of the `checked` axis with the part ABSENT at `unchecked` — the mint's single-axis fit refused with a FALSE receipt ("differ without correlating to any variant axis"); fixed with presence-ragged fill (`partAbsentCombos`, named `PRESENCE-RAGGED` on the binding). (b) the label's base ink equals the control's because Polaris inks the page globally — same family as №5; fixed with the `page-inherited-ink-admitted` door (a control whose ink is not UA black was inked by the library). Case: `presence-hidden-axis-geometry` (WRONG-NAME→PASS, red demoed by disabling the branch). Siblings: polaris.button, radio-button, text-field re-derived with base ink.

**7. fluent.checkbox — square AND circular marks: CORRECT-AS-CAPTURED.** Fluent UI v9's Checkbox genuinely has `shape: 'square' | 'circular'` (microsoft/fluentui#21834; react.fluentui.dev ships the Circular story), and the capture ledger records circular combos against @fluentui/react-components@9.74.5. Documented; no fix.

## Corpus-wide effect — which emissions moved

Every change is attributable to a named door/receipt; non-grid, non-shadcn, non-refused libraries' scripts are byte-identical.

- **fluent**: card, dialog, checkbox, input sets + 00-tokens + GENESIS (contract re-derivation; grid runtime); tooltip's root text now reaches the canvas (its receipt pin retired per its own instruction) and its hug-ceiling pin learned that content wider than the ceiling wraps AT it.
- **shadcn**: select, alert, input (base border colours via the reset door), checkbox (stroke-glyph assets + markup), 00-tokens + GENESIS.
- **astryx**: all ten emitted sets + 00-tokens + GENESIS — the shrink-to-fit refusal plus the ink door re-minted the label ink family, which UNBLOCKED astryx's promote (the reanchor ledger's `imported.button.label.color.primary` exists again; 29/36 acked rows re-applied, 7 stale rows named in REANCHOR-STALE.md for a human).
- **polaris**: all twelve sets (presence-ragged insets, page-inherited ink; generate.ts full 76-surface regeneration).
- **carbon**: inline-notification + GENESIS (it contains the corpus's other aligned grid child).
- **first-party**: the five grid scripts (bentogrid, gridgallery, sidebarlayout, twocolumn, pageshell) + two batches — the aligned-grid-child and margin-box-wrapper runtime fixes.
- **plugin engine receipt** re-recorded (core moved); census code half re-rendered for the 24 changed rows; conformance BASELINE + CANVAS-BASELINE recorded (97 CSS/DOM cases · 94 pass · 3 pre-existing pinned reds; 58 canvas cases · 0 red).

## Red → green

Every case authored BEFORE its fix and captured RED against the stashed engine (verdicts in the run log, quoted): `flex-direction-by-axis` WRONG-NAME, `grid-zero-track-collapse` WRONG-NAME, `border-color-control-equal` SILENT-LOSS, `svg-stroke-glyph-fill-none` WRONG-NAME, `block-root-shrink-to-fit` UNDECLARED-CARRY(HARMFUL), `presence-hidden-axis-geometry` WRONG-NAME (red demoed by disabling the mint branch). All six PASS in the recorded baseline; the canvas round-trip half classifies them ROUND-TRIPPED/NAMED/REFUSED-BY-NAME with 0 red.

## Named walls (not regressions)

- astryx re-promotion: 7 reanchor rows (28 leaves) are FC-THEME-BASE-stale — named in `examples/astryx/tokens/REANCHOR-STALE.md`; a human decision closes them.
- polaris.button's Wave B.4 tone curation (`default` tone) is reproducible by NO committed pipeline — it was grafted back verbatim this round and the gap is named in the closure commit.
- astryx.card now HUGS honestly; the real container-fill width needs a `blockStage: true` recapture (named in the refusal receipt).
- Canvas PNGs are the owner-rejected ones by design: this PR does NOT re-mint to Figma — a later pass re-mints everything from the fixed scripts.

## Gates

tsc · lint · format:check · docs:check · ci:lanes · conformance (CSS/DOM 97: 94 pass + 3 pre-existing pinned reds; canvas kit 157; round-trip 58: 0 red) · census self-test + --phase full · figma:fresh · generated:fresh · maintain — all green at HEAD (full suite recorded in the session log).


tpitre:
Superseded by #62 — `census/stack-to-main`.

This branch's content is merged there on top of `main`, re-derived on the merged engine and measured: **226 of 230 evals** on a clean tree, with the four reds named in `docs/23` §B.38 rather than re-recorded green.

Note for the record: `burn2/landed` was **not** a descendant of `census/canvas-full` — `git merge-base --is-ancestor` returned false, and all 805 real-library reference PNGs were absent from that head. #62 merges #38 explicitly, and the blind re-grade's **117/170** survives intact along with the 471-line PRESERVED LOG.

Closing in favour of #62.

PR #44, opened 2026-08-25
REQUIRED FACTS — a set missing a load-bearing fact refuses to mint and says what is missing

tpitre:
Three held-out subjects, authored **blind**, so the project's new headline metric — first-pass success on a library the tool has never seen — measures the tool rather than the preparation.

## The blindness rule

> The libraries' own documentation and shipped source may be read to author their sandbox and capture config. Our capture / promote / emit chain may **NOT** be run against them. No output of ours for them may be looked at. Nothing may be tuned in response to our pipeline's behaviour.

It is restated verbatim in every README, PROVENANCE, capture config, ds-library manifest, seed contract and token-build script this PR adds. The exam's rules, denominators and exact command list are in [`parity/receipts/v1/HELD-OUT-MANIFEST.md`](parity/receipts/v1/HELD-OUT-MANIFEST.md).

## The three subjects

| # | subject | pinned | components |
|---|---|---|---|
| 1 | **Radix Themes** | `@radix-ui/themes@3.3.0` | **10** — Button · Badge · Checkbox · Switch · TextField · Callout · Card · Avatar · Progress · Tabs |
| 2 | **Bootstrap 5** | `bootstrap@5.3.8` | **10** — Button · Alert · Badge · FormControl · FormCheck · Card · Progress · Spinner · NavTabs · Modal |
| 3 | **react-day-picker** (complex-archetype probe) | `react-day-picker@10.0.1` | **1** — Calendar, 93 DOM descendants |

Every Radix axis value is copied verbatim from the shipped `propDefs`. Bootstrap exports no components at all, so every mount is its own documented markup through `library.customElements` — **no React component was invented**. `examples/day-picker/README.md` §5 spells out, *before any result exists*, seven ways a calendar sits outside the proven archetype list of docs/23 §C.1.1, so its failures are interpretable rather than surprising.

None of the three ships a webfont, so no `fonts` block anywhere; all three stylesheets are `url()`-free, so the harness stays network-free.

## Verified before commit — the library rendering itself, nothing of ours in the loop

- **21/21 components mount and render a root element**, at their default combo and again at their largest enum combo. **0 zero-boxes, 0 console errors, 0 React warnings.**
- Screenshots of every mount in each git-ignored sandbox.
- **21/21 seed contracts parse under `ContractSchema`**; `loadConfig()` accepts all three configs.
- Every committed DTCG name checked against `getComputedStyle` on the live element — **0 VALUE-DIFFERENT** across all 1,256 names (residuals are whitespace inside multi-line box-shadow/gradient strings and CSS quote style inside font stacks).

## Seven findings, banked before the exam ran

1. `importName` cannot name a compound export (`TextField.Root`) — 3 of Radix's 10 need the shadcn-precedent sandbox barrel.
2. **Two class-token axes on one component cannot both be expressed.** For a class-based library both want `className`, and `$props` is last-writer-wins, so Bootstrap's `variant × size` would **silently drop the variant**. `btn-sm`/`btn-lg` deferred by name.
3. **The grammar cannot spell a `Date`**, and a calendar is a function of Dates — unpinned, the DOM changes at midnight. Pinned through a three-literal sandbox fixtures module via `$import`.
4. Child-part axes remain undeclarable (docs/21 §7.3) — new instances named.
5. A component with no steady state has no honest geometry (Bootstrap's spinners read 43×43, then 10×10, then 4×4).
6. `<Theme>` is a real DOM box carrying all 1,091 custom properties with nothing at `:root` — the docs/21 §4.2 trap in its purest form.
7. **A latent defect in two of our own gates.** `drift-check` and `ua-baseline-check` both fell back to the un-namespaced `extract/computed/out/`, which also holds the **first-party** component dirs (`button/`, `badge/`, `avatar/`, `checkbox/`, `textfield/`, `spinner/`). Three configs that have **never been captured** produced eight confident findings about themselves out of another library's scorecards — `bootstrap5/Button: UNPINNED`, `radix-themes/Checkbox: base capture missing`. Fixed by naming the two configs that legitimately use the root; every committed library's resolution is byte-identical and a collision is now an ordinary named skip.

## Existing gates touched — all four handled here

- **`docs:check`** — capture-config count 10 → 13 in docs/21.
- **`census:check`** — three new named exclusions; **denominator unchanged at 170 sets**. Recording the receipt needed one line: `--write-receipt` was wired for `--phase design-to-code` only, so on `--phase code` the gate's own remedy message could not be followed. Check mode still never writes; `--self-test` passes.
- **`drift-check` / `ua-baseline-check`** — per finding 7.

No committed measurement moves, because no capture ran: 116 components still re-fuse.

## Gates — full suite on the finished tree, all exit 0

`typecheck` · `lint` · `format:check` · `docs:check` · `ci:lanes` · `maintain` · `census:check` (+ `--self-test`) · `ua-baseline` · `drift` · `contracts:migrate:check`

## Not created, on purpose

No `extract/computed/out/<lib>/`, no `contracts/`, no `figma/`, no `storybook/`. Each minted tree is a **zero-leaf stub** under the documented `tokens.mintedBootstrap` allowance — these are genuine first-ever passes, and the exam is what fills them.

**Full suite recorded on a clean tree: 225/225 at `e974b0ac`** (`eval:record:check` green).


tpitre:

resolved by picking a side.

PR #46, opened 2026-08-25
THE GRAMMAR CAN NOW SAY WHAT LIBRARIES ARE — compound exports, ordered class tokens, date literals; and a coverage spec for what it still cannot

tpitre:
Preparing the held-out exam **blind** (#43) surfaced seven findings. Three were holes in the **capture config grammar** — the language a user tells us how to mount their library in — not in the engine. They are closed here, **before the first pass runs**, so the exam measures the engine rather than a known gap. A first-pass failure caused by a language hole teaches nothing.

The blindness rule holds: **no held-out subject's component selection or axis VALUES were tuned.** The language changed, not the answers.

---

## 1 · Compound export names — `importName: "TextField.Root"`

`importName` was emitted **verbatim** into `import { … } from '<library.package>'`, so a dotted name was a syntax error. Three of Radix Themes' ten components were unmountable, and the workaround — a sandbox barrel — makes `library.package`/`version` describe *the barrel* rather than the subject.

A dotted name now splits into a **root binding** (imported) and a **member path** (referenced), keyed by the full dotted string so two members of one namespace cannot collapse onto one key.

**Proof.** A synthetic package exporting one namespace object with two members, driven through the real `buildHarnessPage` and real esbuild:

```
IMPORT LINE : import { TextField } from 'fakelib';
COMPONENTS  : const COMPONENTS = { "TextField.Root": TextField.Root, "TextField.Slot": TextField.Slot };
esbuild BUNDLED the page without error
```

The identical config on the parent commit:

```
✘ [ERROR] Expected "}" but found "."
```

**Regression bar.** With no dotted name present the ES6 shorthand is emitted exactly as before — all **13 committed configs generate byte-identical harness entries**, diffed against the parent commit's builder.

**The silent half, also fixed.** `seed-gen.ts` resolved `.d.ts` declarations by name; `propsOf` returns `[]` on a miss, and the props-interface regex read the dot as a **wildcard**. A compound component would have proposed **zero axes**, and per this repo's own doctrine the receipt would have blamed *judgment* rather than the resolver. `declFor` is now compound-aware (flattened member name, then namespace root).

## 2 · Ordered class-token axes — `{"$classTokens": ["btn-primary"]}`

`comboProps` folds every axis into one flat prop bag **by assignment**, so two class-token axes were last-writer-wins. Bootstrap's `variant × size` mounted `className: "btn btn-lg"` and **silently dropped the variant** — no error, no receipt, a plausible-looking capture with an axis missing. `btn-sm`/`btn-lg` was deferred by name rather than ship it.

`$classTokens` **appends** in declared axis order on top of `fixedProps.className`.

**Proof.** Bootstrap Button, un-deferred to `variant(11) × size(3)`:

```
axes: variant(11) x size(3) | combos: 66
distinct className strings: 33
    "btn btn-primary btn-sm"  "btn btn-primary"  "btn btn-primary btn-lg"  …
variant DROPPED anywhere?  false
```

The collisions the append form cannot express are **refused by name at load**: two axes assigning one prop, and `$classTokens` mixed with `$props` on `className`.

## 3 · A date literal — `{"$date": "<ISO>"}`

The marker grammar had `$callback`/`$import`/`$render`/`$element` and no way to spell a `Date`, while a calendar's entire rendering is a function of Dates.

It is a **literal by design** — no `now`, no offset, no local-time spelling — because determinism is the whole reason the construct exists. **V8 does not throw on Feb 31; it returns March**, so a silent rollover is refused by name too.

**Proof.** The emitted resolver, lifted verbatim out of a real harness entry:

```
resolved month  : true 2026-01-01T00:00:00.000Z
resolved today  : true (=== Date.UTC(2026,0,15))
resolved nested : true 2026-01-05T00:00:00.000Z   (markers resolve at any depth)
two resolutions identical (no clock read): true
```

The day-picker sandbox fixtures module is **retired** — the three instants are byte-identical to the `new Date(Date.UTC(...))` literals it held, and nothing hand-written now sits between the config and the subject.

**Eight load-time refusals fire by name.** Three eval cases pin them: `grammar-compound-import`, `grammar-ordered-class-tokens`, `grammar-date-literal`.

---

## The fourth structural artifact

`spec/GRAMMAR-COVERAGE.md` + `spec/grammar-coverage.json`, in the family of the channel table and the door register.

**44 supported and 10 unsupported constructs**, 54 rows, measured against 13 committed configs using 89 distinct construct paths.

Every unsupported row names the library that proves the need and the cost being paid **now**, not a hypothetical:

| construct | proven by | the cost today |
|---|---|---|
| `child-part-axes` | bootstrap5, radix-themes, carbon | the child is pinned and the axis deferred by name; the contract cannot say how many renderings are missing |
| `no-steady-state` | bootstrap5 | the grow spinner reads 43×43, then 10×10, then 4×4 — whatever frame the sampler caught is minted as geometry |
| `theme-provider-is-a-box` | radix-themes | 1,091 custom properties on `.radix-themes`, nothing at `:root`; omit `wrapperOpen` and every value is minted as a literal, silently |
| `boolean-variant-axis` | day-picker | `showOutsideDays` is a boolean that *selects a rendering*; `axes` refuses non-enums and `stateProps` refuses non-states — it falls between them |
| `multi-package-mount` | radix-themes, shadcn | still costs the Radix barrel; deliberately **not** retired here, because that would change a held-out subject's mount |
| …and five more | | |

**The gate measures the supported half from the CONFIGS ON DISK, never from `ComponentConfig`** — a coverage number computed from the declaration that defines its own denominator is the self-attestation this tree keeps re-learning. It refuses when a config uses a construct no row claims, and when a row claims support that nothing exercises. `--self-test` plants **ten** reds, including an empty config corpus, and requires the committed spec to be green first.

Wired into the **fast lane**; `npm run ci:lanes` green.

---

## Merge note

This branch merges `exam/held-out-prep` at `ac9974cf`, which fixed the **third** instance of finding 7's root-fallback bug in `scripts/door-census.ts`. I had independently fixed the same bug while landing this work; on merge I **took `ac9974cf`'s version wholesale** and dropped my duplicate, so this PR carries no competing refactor of that fix. The manifest annotation credits `ac9974cf`.

---

## One inherited red — fixed on the exam branch, not here

The suite first came back **227/228** on the merged tree: `gate-inventory-shipped-minted` banned `tokens.mintedBootstrap` outright, while `loadConfig` has supported it since task #28 for a library's genuine **first-ever** pass. The exam branch's three held-out subjects — committed complete, captured by nothing, deliberately — carry the flag, so the eval and the loader contradicted each other. The red was inherited, not introduced here: the flag arrives with `exam/held-out-prep`, and `main` had no such configs.

I fixed it, and `exam/held-out-prep` fixed it independently at `f3e4400e`. **On merge I took the exam branch's version and dropped mine**, so this PR carries no competing edit — the same resolution I applied to `door-census.ts` and `ac9974cf`. Both crossings stay refused by name: a shipped tree carrying the flag, and a zero-leaf tree not declaring it.

## Merge state

Merges `main` at `c4127fc9` (which now includes the first-pass exam, #42) and `exam/held-out-prep` at `f3e4400e`. Two conflicts, both places where the exam branch and I had independently fixed the same inherited bug; **the exam branch's version won both times.**

## Gates

`tsc` · `lint` · `format:check` · `docs:check` · `ci:lanes` · `grammar-coverage:check` + `--self-test` · `first-pass:check` · `conformance` · `census:check` (`--phase code`, `--self-test`, `--phase design-to-code`) · `figma:fresh` · `generated:fresh` · `maintain` · `channel-table:check` · `door-register:check` + self-test · `capability:fresh` · `eval:record:check` · `eval:registry:check` — **all exit 0.**

**`npm run eval` — 228/228**, recorded on a clean tree at `15b67110` (`dirty: false`), which is the merged head rather than a commit behind it. The suite grew 225 → 228; by claim family the additions are 2 refusal + 1 determinism, so docs/22's breakdown moves 36→38 refusal and 35→36 determinism. Every doc number quoting the suite size is re-derived, and `docs/24-what-works.md` is regenerated by `npm run capability:report`.


tpitre:

resolved by picking a side.

PR #47, opened 2026-08-25
THE DOCUMENTED PATH NOW REPRODUCES ITS OWN CORPUS — 7 libraries re-derive from committed inputs; the undocumented prune is gone

tpitre:
The first-pass exam (#42) measured the metric this project had never measured — **does the documented chain work on the first attempt, untouched?** — and got **0 of 8**. Two of the three causes were ours. Both are fixed at the cause, and each now has a gate that refuses when it comes back.

## What moved

| exam | before | after |
|---|---|---|
| `selftest-flowbite-live` (canvas→code) | 4/8 chains complete | **8/8**, first try, no prune |
| `selftest-tailwind` (code→canvas) | 0/8 — **all eight died at `promote`**, stage 2 of 6 | 0/8 — all eight run `capture → promote → validate → generate → bundle` clean and stop at **`mint`**, which is MCP-driven and not a capability this harness has |
| `selftest-altitude` (code→canvas) | 0/8 — stopped at mint | 0/8 — stopped at mint (unchanged; re-run) |

The ratchet now records **rises and shape-moves**, not only falls — a number that quietly rises is not evidence either.

## Finding 1 — the corpus was not re-derivable from its own inputs

A fresh capture of flowbite `Alert` produces props `[color, children]`. The committed capture record carried `[color, icon, dismissable, children]` **plus six anatomy parts**, and no committed seed, config or sandbox produces them. `ac5e6181`'s own commit message says where they came from: *"the hill-climb hand-edited committed contracts … without back-porting promote inputs"*. So `promote` refused every tailwind set by name, and the exam scored 0/8 before reaching a single interesting stage.

Those four facts are **inputs** now. The authored-facts door gains an **`add`** operation — whole props and whole anatomy parts, placed before/after a named sibling, refusing BY NAME the day the capture learns to carry them. All 24 tailwind artifacts still re-promote byte-identically, and a fresh capture through the documented chain now promotes clean.

**Measured across the whole corpus, both derivations, for the first time:**

| library | promote (record + authored → contracts, bytes) | capture (seed + config + sandbox → record, structure) |
|---|---|---|
| altitude | ✔ 25 | 5/8 — 3 hand-edits from `c924c9c2` (FC-ENUM-HOLE ×2, a hand-authored `label` part) |
| antd | ✔ 37 | PENDING — its sandbox is git-ignored and not on the measuring machine; never counted as reproducing |
| carbon | ✔ 30 | 8/10 — 2 engine-drift rows (a `::before` the reader learned later; Modal's root selection moved) |
| fluent | ✔ 33 | **11/11** |
| mui | ✔ 83 | 28/31 — a prop from nowhere (`snackbar.position`, citing a config that never existed), one engine-drift row, and `tooltip` |
| polaris | **✘ 9** — hand-edited contracts, no authored ledger | **12/12** — its records re-derive perfectly; the break is entirely record→contract |
| shadcn | ✔ 32 | **11/11** |
| tailwind | ✔ 26 | **11/11** (was 10/11) |

**The two derivations fail in different places, which is why they are two gates.** polaris re-promotes badly and captures perfectly; carbon and mui re-promote perfectly and diverge on capture, where their records sit behind engine fixes that landed after they were last recorded. One combined "is the corpus reproducible?" number would have hidden both.

The sharpest row is **`mui/tooltip`: the capture refuses itself.** The Popper's x offset is `19px` on one sweep and `16px` on the next, so `extract/computed/run.ts`'s own double-run determinism check stops — that record cannot be re-derived by anyone, on any machine, today. Honest refusal, now on the record.

**New gate:** `npm run corpus:reproducible:check` (fast lane, ~2 s, no browser) · `-- --capture` (out of band; it **records its measurement** so the verdict is identical on every machine) · `-- --self-test`. Every divergence must be named in `parity/receipts/v1/corpus-reproducible.json` with its reviewed cause; an unnamed one refuses, and **so does a named one that has since been fixed**.

## Finding 2 — the documented path needed an undocumented repair

Both dump producers round canvas geometry to two decimals (`round2` in `dump.plugin.js` and `rest/map.ts`). So a width the code→canvas mint spelled `39.9219px` from computed style comes back `39.92px` — **one measurement, two spellings, on one token path** — and `generate` rightly refuses a two-valued slot. The census healed it with a token prune the CLI has no flag for; the exam, which performs no repair, took the refusal.

Fixed in the minter: `MintOptions.corpusValueAt` asks the corpus what it already spells at each claimed path, and `sameUnderDumpRounding` carries the corpus spelling when the observation is that value re-rounded. Six leaves reconciled, each **named** in the proposal report; zero conflicts remain.

And the census's prune is now a **checked de-dup** that REFUSES a value disagreement instead of resolving it in its own favour. The internal path no longer has a step the user's path lacks.

## Finding 3 — mint is an MCP-driven stage, modelled as one (not fixed)

`probeBridge()` scanned `127.0.0.1:9223-9232` for a "COMMAND endpoint" and printed what answered. The refusal was honest; the premise was not — **there is no such endpoint.** Printing `9228:200` invited the reader to believe the write was one configuration away. It is not one configuration away; it is a different actor.

Every `StageRecord` now carries `driver`. `mint` is the one stage with `driver: "mcp"`, and its evidence is a file an MCP-holding agent writes (`<packet>/mint-evidence.json`). No evidence, no mint: PENDING with the architecture named, never a capability implied. **The no-retry rule is unchanged and now binds the agent too.**

## Also — committed example bundles could go stale with nothing noticing

Only four of the eleven `examples/*/figma/*.bundle.json` — the ONE JSON a designer pastes — were ever rebuilt and byte-compared. **New gate `npm run bundles:fresh`** (full lane) rebuilds every one with the command its own PROVENANCE documents. **Three were stale**: polaris, astryx and astryx-docs were missing the `codeOnlyFacts` receipts the engine had learned to write (~4.2 KB each; ~9.5 KB for polaris). Executing a stale bundle succeeds, which is exactly why nothing noticed. Rebuilt and committed. `astryx-genesis.bundle.json` is FROZEN on its sha256 with the reason on file — a version-1 contracts-only envelope for a flow that does not exist yet, which no documented command re-derives.

## What this PR does NOT fix, by name

- **polaris's 9 promote divergences** and **altitude's 3 capture divergences** — the authored-facts door can carry all of them; routing them is a bounded mechanical round of its own. Named in the ledger, machine-checked, not silent.
- **mui/snackbar's prop from nowhere** and **mui/tooltip's non-deterministic capture** — same, with the cure spelled out in each row.
- **carbon and mui's engine-drift rows** — the records are BEHIND the engine, not the other way round; the cure is a recapture round that moves those libraries' scorecards.

## Gates

`tsc` · `lint` · `format:check` · `docs:check` · `ci:lanes` · `first-pass:check` + self-test · `corpus:reproducible:check` + self-test · `bundles:fresh` · `door-register:check` · `channel-table:check` · `conformance` · `figma:fresh` · `generated:fresh` · `maintain` · `plugin:check` · `plugin:ui-check` · `exact-proposal:check` · `first-party-bundle:check` · `code-only-facts:check` · `emitters:check` · `mint:check` · `census:check` (`--phase code`, `--phase design-to-code`, `--self-test`) — all exit 0 (26 gates). `npm run eval` **225/225**, recorded on the CLEAN tree.

Merged `origin/main` (channel table + door register). The two conflicts were pure additions on both sides and both sides are kept; the 26 `@door` line pins in `core/mint-tokens.ts` were re-recorded from the markers themselves, with each door's `ruleLine` shifted by the same delta. No door added, removed or changed: 425 named, 164 still silent, census unmoved.


tpitre:

resolved by picking a side.

PR #48, opened 2026-08-25
AN EXAM THAT CANNOT RUN MUST NOT DESTROY THE EVIDENCE — preflight before clearing, and 'measured nothing' never exits 0

tpitre:
## The defect, reproduced before anything was changed

`npm run exam:first-pass -- --exam <name> [--mint]` called `clearPacketImages()` for the whole exam **before the chain proved it could run**.

In a fresh worktree the git-ignored library sandbox (`examples/altitude/.altitude-sandbox`) does not exist — no clone or worktree has it — so all 8 sets died at `capture` with *"need --harness <dir> with altitude-web-components@1.0.2, react@18, react-dom@18, esbuild installed"*. By then the committed packet was already gone:

```
$ npm run exam:first-pass -- --exam selftest-altitude
  ✖ Button   ERROR at capture   0.7s  ref 0 · code 0 · canvas 0
  … ×8
  FIRST PASS: 0/8 chains complete on attempt #1
$ echo $?
0
$ git status --short parity/receipts/v1/first-pass/selftest-altitude | awk '{print $1}' | sort | uniq -c
  94 D          # every committed ref/code PNG deleted
   9 M          # 8 attempt.json + MANIFEST.json, rewritten as an ERROR-at-capture packet
```

**A run that could not possibly measure anything destroyed the evidence of the run that could, and reported success to the shell.** `first-pass:check` afterwards complained only that the receipt was stale — it accepted the wiped packet, because every wiped surface came with an `absent` reason.

### 1. Preflight before any destructive step

`preflight()` (`extract/figma/census/first-pass.ts`) validates every precondition and **refuses by name with the exact command that fixes it, before a single committed byte is touched**:

- the declared capture harness exists, carries the library at the version the capture config pins, and has `react`, `react-dom`, `esbuild`;
- the capture config and `ds-library.json` exist and parse;
- direction B: a Figma token is present, the file key is not forbidden, every declared corpus file is on disk (an absent one used to be silently dropped — a substituted input), at least one page is selected;
- work dir and packet dir are writable.

```
$ npm run exam:first-pass -- --exam selftest-altitude
✖ PREFLIGHT REFUSED — exam "selftest-altitude" cannot run, and NOTHING was touched.
  The committed packet under parity/receipts/v1/first-pass/selftest-altitude/ is
  evidence of the last run that COULD measure. A run that cannot produce a
  replacement never clears it.

  ✖ the capture harness (sandbox) examples/altitude/.altitude-sandbox exists
      found: examples/altitude/.altitude-sandbox does not exist — it is git-ignored,
             so a fresh clone or worktree never has it
      fix:  mkdir -p examples/altitude/.altitude-sandbox && cd examples/altitude/.altitude-sandbox \
              && printf '{"name":"altitude-sandbox","private":true}\n' > package.json \
              && npm i altitude-web-components@1.0.2 react@18 react-dom@18 esbuild
$ echo $?
2
$ git status --short parity/receipts/v1/first-pass/ | wc -l
0
```

The remedy is not hand-written prose: it is rebuilt from the capture config's own `library.package` / `library.version` pins, so it cannot drift from what the run actually requires.

### 2. Exit codes tell the truth — "measured nothing" is not "measured zero"

`isMeasured()`: a set is MEASURED when the **pipeline** produced its outcome — a completed chain, an engine `REFUSED` **by name** (the finding the exam exists to collect, and `0` is the right exit for it), or a `PENDING` stage the harness does not execute at all (`mint` is MCP-driven). A set that stopped at `ERROR` measured nothing.

| exit | meaning |
|---|---|
| `0` | the run measured something, including an honest refusal |
| `1` | every set stopped at `ERROR` — **UNMEASURED**. No rate is recorded anywhere |
| `2` | preflight refused; the run never started |

An UNMEASURED exam renders as `UNMEASURED` in the receipt table, not as a number; `applyRatchet` records nothing for it (not a new row, not a raise, not a lowering); and `first-pass:check` refuses it by name. `0/8` says *the engine failed eight times*; the truth was *the harness never got to ask*. Collapsing those is the same class as the killed-suite-reads-as-pass defect this repo already paid for.

### 3. Destruction is narrow and recoverable

`clearPacketImages()` is now called **per set, at the instant that set's replacement is written**:

- direction A: the copy plan is built from the shadow work dir first; the clear happens only when the plan is non-empty;
- direction B: the up-front bulk clear of the whole exam is gone; the clear happens inside `renderGenerated` immediately before the first screenshot of that set lands;
- `exportRefs` now removes each `ref-*` target before fetching, so a leftover PNG can never be counted as this run's Figma render (`existsSync(outPath)` is that stage's only evidence).

A set that never reaches the point of replacement keeps every byte and records it as `images.retained`, with the reason in `images.absent`. The receipt says out loud that the pictures beside that attempt are older than it.

Verified end to end with a fabricated sandbox that passes preflight but cannot capture:

```
  8 set(s) produced no image and therefore CLEARED NOTHING — their committed images
  are from an earlier run and are recorded as retained: Button, Badge, Chip, …
  ✖ MEASURED NOTHING: all 8 set(s) stopped at ERROR … 0/8 is NOT a rate
$ echo $?; find … -name '*.png' | wc -l
1
94        # every committed PNG intact

$ npm run first-pass:check
✖ UNMEASURED: exam "selftest-altitude" attempted 8 set(s) and MEASURED NONE …
$ git diff --stat parity/receipts/v1/first-pass-ratchet.json   # after --write-receipt
(empty — the ratchet did not move)
```

### 4. The gate refuses the new ways to lie

- **per-SURFACE blank cell**: no `ref` image + no `ref` absence + no retained ref is red on its own. The old whole-packet rule could be satisfied by a single `canvas` absence while `ref` and `code` had been silently wiped — which is exactly how the defect produced a packet the gate accepted;
- **orphan image**: a PNG the attempt names as neither its output nor retained evidence;
- **unmeasured exam** trying to carry a rate.

Self-test grew from 5 planted reds to **8**, plus **6 preflight red cases** — including one that proves the guard is not a blanket no (a complete sandbox must not refuse) and one for library version drift.

## The same class, found elsewhere

Grepping the exam/census paths for a clear over committed evidence turned up two more:

- **`extract/figma/census/render.ts`** `renderRow()` removed every committed `code-*.png` for a row at the **top** of the function — and the next statement after it can be `return receipt` with `unavailable = "engine compile refused"`. A row the engine declined lost its committed renders and got nothing back. Deferred to the first raster write.
- **`extract/figma/census/d2c-render.ts`** removed every committed `code-*`/`canvas-*` PNG for a set before the screenshot loop, whose **first** statement can `throw` on a cell that did not mount, leaving the pair directory wiped and nothing written back. Deferred to just before the first screenshot lands.

Neither changes behaviour for a run that succeeds — the clear still happens once per row/set, before the first new byte. Only the failing path differs, and only by keeping what it can no longer replace. The census gate checks that every PNG a receipt *names* is on disk, never the reverse, so retained rasters beside an `unavailable` receipt are gate-safe.

Checked and **not** of this class: `design-to-code.ts:325` and `corpus-reproducible-check.ts:797` remove scratch work dirs; `canvas-census-check.ts:530` plants a red inside its own staged copy; `first-pass.ts` `makeShadowRoot` removes a symlink tree in `$TMPDIR`.

## Gates

All run serially on the clean tree.

| gate | result |
|---|---|
| `tsc --noEmit` | 0 |
| `lint` | 0 |
| `format:check` | 0 |
| `docs:check` | 0 — 41 documents gated |
| `ci:lanes` | 0 |
| `first-pass:check` | 0 |
| `first-pass:check --self-test` | 0 — 6 preflight red cases + 8 planted reds, all named |
| `census:check --self-test` | 0 |
| `census:check --phase code` | 0 — 170/170 |
| `census:check --phase design-to-code` | 0 — 23 sets, 0 SILENT |
| `maintain` | 0 |
| `eval` (full suite) | 0 — **225/225**, re-measured on the clean tree at `7ed9e2da`, `dirty: false` (self-verified: the record's `commit` equals this branch's HEAD, not an inherited one) |

The bare `census:check` (default `--phase full`) is red on this branch for the pre-existing reason it is red on the base — the canvas half of the census has no minted PNGs or verdicts yet. CI runs the two phases above, both green. No census receipt byte is touched by this branch.

## One more instance, and one regression I caught in my own guard

- **`runCanvasToCode` wrote `MANIFEST.json` unconditionally.** If the page selection came back empty, it would write a MANIFEST naming **zero sets** over the committed one, and every existing packet directory would instantly read as SMUGGLED — the images intact, the record of which sets the exam *has* gone. It now refuses before that write; the runner catches a refusing runner and exits 1.
- **Preflight's writability probe would have created the packet directory it was probing.** `writable()` mkdirs recursively, and it was called on `<packet dir>/<exam>` — so for an exam with no packets yet, a *refusing* preflight would leave an empty exam directory behind: a packet with no MANIFEST, exactly the red the gate exists to catch. Git does not track empty directories, so "nothing was touched" would have been true of the diff and false of the gate. It probes the packet tree root instead, and the self-test asserts a refusing preflight creates no directory of its own.

## No behaviour change for a healthy run

`selftest-tailwind`, `selftest-altitude` and `selftest-flowbite-live` all remain MEASURED (they stop at `mint` PENDING or complete), the ratchet is byte-identical, and the only receipt change is the new prose explaining `UNMEASURED`.


tpitre:

resolved by picking a side.

PR #50, opened 2026-08-25
THE MINT STAGE IS DRIVEN — code→canvas completes end to end for the first time, with the canvas evidence in the packet

tpitre:
`mint` has been PENDING since the first-pass exam existed. The harness cannot drive it — the figma-console bridge speaks MCP over stdio to its own client and WebSocket to plugin clients, and a Node process is neither — so every code→canvas exam stopped at *"bundle produced, mint pending"* and the direction had **never been measured past `bundle`**.

It has now been driven, once, with evidence. **`selftest-altitude`: 0/8 → 7/8 chains complete, 5 minted, 30 canvas PNGs in the packet.**

---

## What ran

`npm run exam:first-pass -- --exam selftest-altitude --work <kept> --mint` on the clean tree at `dde1d4f4`. Preflight green. All 8 sets clean through capture → promote → validate → generate → bundle; all 8 stopped at `mint` PENDING. Exit 0.

Then an agent holding the figma-console MCP tools drove the write. Each emitted `mint/<component>.figma.js` was served on `http://localhost:9232`, fetched **inside** the plugin sandbox, **sha256-verified against the hash the harness had recorded for those same bytes**, and evaluated **exactly once**. No retry, no repair, no hand-edited script, no fixed-up mint.

| set | outcome | node | detail |
|---|---|---|---|
| Button | **minted** | `39:6719` | amended, 12 variants rebuilt, 4 reactions wired |
| Badge | **minted** | `39:6692` | amended, 8 variants rebuilt |
| Chip | **minted** | `39:6746` | amended, 5 rebuilt, 5 focus-visible added, **5 `Type=Default` left behind** |
| Link | skipped — unchanged | `39:6823` | specHash already equal; **no byte moved** |
| Avatar | **minted** | `39:6673` | amended, standalone component |
| Heading | skipped — unchanged | `39:6778` | specHash already equal; **no byte moved** |
| Divider | **REFUSED** | `39:6751` | *"set/standalone shape mismatch (COMPONENT_SET vs isSet=false) — a human retires the old node"* |
| IconClose | **minted** | `39:6802` | amended, 7 variants rebuilt |

## The finding, which is not in the numerator

**Not one of the 8 took the CREATE path.**

The operator created an empty page `first-pass altitude 2026-08-25`, cleared it, and made it the current page before every run. All 8 scripts ignored it — `newPages` was `[]` every time and the page was still empty at the end.

The emitted script owns file layout. `resolveComponentIdentity` scans **every page in the file** for a node stamped with the same `ds_contracts/contractId` and amends it in place; only when nothing matches does it create, and then on its own page named after the component. There is no `--page` flag, no file-scoping, no documented way to confine a mint to one page. The Scratch Project is the only writable file and already carried all 8 altitude stems from the canvas census — so in it, a first-pass mint can only ever measure **amend**.

Chip shows the cost: the amend left 5 `Type=Default` variants the contract does not carry standing on the canvas (it reported them as `extraVariants`), so the set now mixes two vocabularies.

`7/8` is a true statement about the documented chain and a **weaker one than it looks**. The ratchet row, `mint-evidence.json` and docs/31 §6.3 all say so in those words. The one door that would change it — `globalThis.DS_CREATE_ONLY` — is a pre-run switch of the artifact, i.e. exactly the substituted input the metric forbids, so an exam does not set it.

## `minted` is deliberately narrower than `chain complete`

A mint the engine skipped as unchanged did its job and completes the chain, but **this attempt moved no byte**. `StageRecord.bytesWritten` carries the difference, the tally counts only true writes, and the receipt legend says which is which. A number labelled "minted" may not quietly include "we wrote nothing" — so the header row reads `chain complete 7`, `minted 5`.

## The recorder refuses whole

`extract/figma/census/first-pass-mint.ts` (`--record-mint`) folds `mint-evidence.json` into the packet **through the exam's own writers**, so the receipt stays byte-stable. It runs no stage and clears no image, and it writes **nothing at all** if any of these hold:

- a non-scratch or forbidden file key, anywhere in the evidence bytes
- a MANIFEST set the evidence does not name, or a name the MANIFEST does not have
- a `scriptSha256` that is not the one the chain emitted — a hand-edited script can never be recorded as a mint
- a `minted` set with no picture of it
- `bytesWritten` contradicting the status
- a cell with neither a canvas image nor a named reason
- a mint recorded onto a chain that never reached the stage
- **a stage that is not `PENDING`** — which is what stops `--record-mint` from ever becoming a retry

## Documentation

docs/31 §6.1 is the operator procedure, written only after there was real evidence to describe: what an agent runs, the mandatory asserts as a table, and what evidence must be saved. It names the two traps this run hit — the Desktop Bridge's `networkAccess` allowlist is `http://localhost:9223-9232` and **`127.0.0.1` is not in it** (`Error: Failed to fetch`), and the plugin sandbox has `eval` and `fetch` but **no `crypto`**. It also records that a transport failure *before* `eval` is **not** a mint attempt (nothing reached the canvas), while a throw *inside* `eval` is. §6.2 is the five statuses; §6.3 is the isolation finding.

## Not graded

Per the exam's own rule, **this PR does not grade**. No `verdict.json` was written. The 30 `canvas-*.png` sit beside the `ref-*`/`code-*` pairs under the packet's own cell names, ready for a separate blind pass.

Every cell with no canvas image is named with a reason — most are the engine's own `FC-UNSET-PLANE-UNDRAWN`, which it reports in its own code-only facts. Divider's two cells carry the refusal instead: its node on the canvas is an **earlier** mint's output, and exporting it would pass another run's pixels off as this attempt's.

## Gates

`tsc` · `lint` · `format:check` · `docs:check` · `ci:lanes` · `first-pass:check` + `--self-test` · `census:check --self-test` · `maintain` — all exit 0, serially, on the clean tree. `first-pass:check --self-test` still goes red on all six planted lies.


tpitre:
Subsumed by #53 — https://github.com/southleft/ds-contracts-poc/pull/53

fix/mint-doubled-text-and-ink (#51) descends from this branch (`git merge-base --is-ancestor exam/first-pass-mint fix/mint-doubled-text-and-ink` is true), so merging #51 into the consolidation branch carried every commit here: the driven mint stage, selftest-altitude 0/8 -> 7/8 chains, the 30 canvas PNGs, and StageRecord.bytesWritten. Nothing was dropped. Closing in favour of the consolidated PR.

PR #51, opened 2026-08-25
A text-indent that evicts the label is not an annotation — the canvas was inventing ink (and the button ink was correct all along)

tpitre:
### 1. `altitude.badge` — not a doubled label. An **invented** one.

The right-hand column of the minted set is the `Dot=Dot` plane. altitude renders that
plane as an 8px status pip and hides the slotted label in CSS:

```css
.al-c-badge.al-is-dot { padding: 0; min-width: 8px; max-width: 8px; text-indent: 9999px }
```

The minted cell drew the label anyway, at indent 0, inside an 8px frame — so a ~37px
"Badge" overflowed the pip and ran across its neighbours. At 1.5x in Figma that reads as
a dark second copy offset from the coloured one. The committed receipts carry it:
`canvas-info-dot.png` / `canvas-success-dot.png` paint the word; `ref-info-dot.png` /
`ref-success-dot.png` are the dot alone.

**Cause chain**

| layer | what it did |
|---|---|
| capture (`extract/computed`) | measured `text-indent: 9999px` **and** the used width `8px` on that plane, carried the indent as a token, and never asked the question the two numbers answer together |
| contract | root carries `content: {prop: children}` + `text-indent` per axis value; nothing says the browser paints no text there |
| emitter (`core/emit-figma-script.ts`) | `TOKEN_CHANNELS` classes `text-indent` **`annotate`** — Figma text nodes have no first-line indent — so the offset is dropped and *named*, and the label is kept |

An annotation is the honest answer only while the canvas still draws **what the browser
draws**. Here the browser draws *nothing in the box*, so dropping the offset and keeping
the label does not lose a fact — it invents one.

**Fixed at the cause.** A new MEASURED field `Part.textOutOfBox` (never hand-authored —
the `hugsBelowMaxWidth` discipline). `extract/computed/fuse.ts textOutOfBoxEvidence`
asks, per part and per enumerated combo, whether the indent lands at or past the
content-box end edge (`width` minus padding and borders under the element's own
`box-sizing`). Evidence is carried only when the indent, width, padding and borders are
all real pixels; the indent is **positive** (a negative indent leaves the box only if it
exceeds the text's own advance, which the capture does not measure); and the off-box
combos are all of them, or exactly one enum axis's **declared** value subset. Everything
else is a named receipt and no field — previous behaviour, unchanged.
`core/emit-figma-script.ts` then draws no text child on exactly those combos, through
all three label paths (`rootTextSpecs`, the bound-`children` branch, `partToSpecInner`);
the box still paints. `validateContract` refuses evidence that qualifies nothing.

No component name is special-cased and no tolerance moved.

### 2. `altitude.button` — a **false premise**, refuted by measurement.

Every minted button cell is pixel-identical to the real-library reference beside it
(`canvas-*.png` vs `ref-*.png`, same dominant colours to the byte). The "wrong ink" and
the "impossible khaki" are altitude rendering itself correctly:

```
.al-c-button          background-color: --al-theme-color-background-primary-default   -> --al-color-brand-blue-500  #4375ff
                      color:            --al-theme-color-content-primary-weak         -> --al-color-brand-blue-900  #000b29
.al-c-button--danger  background/color: brand-red-500 #f05735 / brand-red-900 #1f0600
.al-c-button--secondary  background/color: brand-taupe-500 #a49981 / brand-taupe-900 #191306
```

(`altitude-web-components@1.0.2`, `dist/components/button/button.js` + `dist/css/main.css`.)
The library's filled variants really do use a same-hue `-900` ink, and its `secondary`
really is **brand taupe** — the "khaki". Sampled pixels match those hexes exactly.
**Nothing fixed; the premise dies on the measurement.**

---

## Conformance / evals — RED before, GREEN after

`evals/run.ts` case `text-indent-off-box-label-not-drawn`, emitter at the pre-fix commit:

```
✖ C3-detection  text-indent-off-box-label-not-drawn
    Error: Dot=Dot DREW the off-box label — the canvas invented ink the browser never paints: ["Sample"]
```

same case, fixed:

```
✔ C3-detection  text-indent-off-box-label-not-drawn
  no evidence ⇒ both planes keep their label; MEASURED textOutOfBox ⇒ Dot=Dot draws no
  text child and the text-indent code-only fact names the withheld label on that plane
```

Plus `text-indent-off-box-evidence-refuses-when-inert` (C2-refusal): 5 inert or
ill-formed spellings refused by name, the measured spelling passes.

New CSS/DOM conformance case `text-indent-off-box` (real Chromium, the unmodified
capture pipeline, then the mock canvas). The **capture** now measures it end to end —
the case's enriched contract carries `textOutOfBox: {}` beside its `text-indent` token —
and the dumped variant changes:

```
before  "children":[{"name":"label","type":"TEXT","text":{"characters":"Sample",…}}]   bbox 8x17
after   "children":[]                                                                  bbox 8x1
```

## Sibling sets

Only altitude **Badge** and **Avatar** carry a `text-indent` channel anywhere in the tree
(`imported.badge.root.text-indent.{default,dot}` and `imported.avatar.badge.text-indent`).
Both sets' `text-indent` code-only-fact reason changed, so their scripts plus altitude's
`GENESIS-BATCH.figma.js` and `altitude.bundle.json` are re-emitted by their documented
remedies (`scripts/reemit-altitude-figma.ts`, the `ds-library.json` bundle block,
`examples/altitude/scripts/build-genesis-batch.mjs`). **No drawn byte moves**: no
committed contract carries the new field yet.

## A second finding, reported not fixed: a hand-edited capture artifact

`extract/computed/out/altitude/badge/enriched.contract.json` is a *measured* artifact,
but commit `c924c9c2` (2026-08-08) hand-edited it: it deleted the measured
`root.content` and wrote in a `parts.label` with `visibleWhen: {prop: dot, equals: default}`
and a prose description explaining that "Figma has no text-indent twin, so the label is
compile-dropped when Dot=Dot". No capture produces that shape — a fresh capture of the
same component emits `root.content` and no parts at all (verified: re-running
`extract/computed/run.ts` reproduces every other byte and never that hunk).

That fabrication is why this defect stayed invisible: the committed contract, bundle and
figma script all inherit it and are *correct*, so `figma:fresh` / `bundles:fresh` stayed
green — while the first fresh chain (this exam) lost it and painted the label. It also
invents a `<span>` the real DOM does not have.

I did **not** rewrite it here. The documented remedy is a re-capture, and this repo's
Chromium has moved 149.0.7827.55 -> 151.0.7922.34 since that artifact was recorded
(795 -> 798 enumerated channels on this component), which would take `channel-table:check`
red on newly enumerated properties and churn pinned numbers across an unrelated surface.
That is an owner-scale event, not this round's. With the engine fixed, the honest
re-capture now produces `textOutOfBox` on its own and the fabrication becomes unnecessary.

---

## Gates, serial, on the clean tree

`typecheck` · `lint` · `format:check` · `docs:check` · `ci:lanes` · `conformance` (92 cases,
89 pass / 3 pre-existing recorded reds, no drift) · `conformance:roundtrip` (55 cases,
0 SILENT / 0 DRIFTED / 0 HARMFUL) · `extract/figma/conformance` · `census:check --self-test` ·
`figma:fresh` (9/9 byte-fresh) · `generated:fresh` · `bundles:fresh` (10/11 + 1 frozen) ·
`schema:fresh` · `channel-table:check` · `capability:fresh` · `maintain` — **all exit 0.**

**`npm run eval`: 227/227.** `evals/results.json` carries `commit dfb652cf`, `dirty: false`,
227 rows, zero failing — measured on that clean tree, with the record itself the only commit
after it (`eval:record:check` and `eval:registry:check` both green).

The suite grew 225 → 227 with the two new cases, so every gated eval-count claim moved with
it (docs/24 re-derived by `capability:report`; prose in README, docs/00, 07, 08, 12, 14, 22
and the four handoff pages).

**`census:check` (full) stays red** — canvas 0/170, no verdicts. Untouched by this branch:
`git diff origin/exam/first-pass-mint..HEAD -- parity/ extract/figma/census/` is empty. The
gate this round owns, `--self-test`, is green, and no `verdict.json` was written.


tpitre:
Consolidated into #53 — https://github.com/southleft/ds-contracts-poc/pull/53

Merged there onto the new main (828d3e26), together with #50 (which it descends from) and #52, so all three land in one lane cycle. The measured Part.textOutOfBox fix is intact; the merge also closed two latent reds this branch carried on its own: part.textOutOfBox was never mapped in site/src/coverage.ts (schema:fresh red, the guard site:build refuses on), and four @door markers in extract/computed/fuse.ts were never registered in spec/door-register.json. Both are fixed in #53. Closing in favour of the consolidated PR.

PR #52, opened 2026-08-25
Pin the render browser, and make every receipt name the browser that produced it

tpitre:
A stray Chromium on one machine rendered every local recording, and no receipt could say so. This pins the browser, makes receipts name their renderer, and removes the two artifacts the stray produced.

**Found while fixing PR #49's only red.** `chromiumExecutable()` resolved the ms-playwright cache by **highest revision present**. A dev machine carried a stray `chromium-1234` (Chromium 151.0.7922.34) for an unrelated tool, while `playwright-core` pins `chromium-1228` (Chromium 149.0.7827.55) — what CI installs and what the committed captures were taken on. Measured cost: 37 drift rows moved, 88 findings in the full lane, and a `darwin` platform baseline recorded on the wrong binary.

---

### 1. The resolver resolves the *pinned* revision, or refuses by name

Revision read from `playwright-core`'s own `browsers.json` (resolved beside its `package.json`, which is in the exports map). No highest-revision sort, **no silent system-Chrome fallback**. `PLAYWRIGHT_CHROMIUM_PATH` still wins — that is an operator naming a binary on the record. The refusal names the pin, what *is* present, and `npx playwright install chromium`.

Not only a dev hazard: the workflows cache `~/.cache/ms-playwright` with `restore-keys`, so a CI cache accumulates revisions across lockfile bumps — "highest present" is wrong there too the moment playwright is downgraded.

### 2. Every receipt names the browser that rendered it

`Scorecard` gains `renderedBy`, and a `browserMismatch` sentence naming both when it disagrees with the capture. **Not a reassignment:** `browserVersion` still carries the capture's browser because it feeds `bareBrowser` into the enriched contract's `description`, which must stay byte-identical on an offline re-fuse. Two different facts, both recorded.

It caught a real case on its first run — `tailwind/Spinner`: `capturedBrowser: 151`, `renderedBy: 149`, mismatch note attached. Previously it printed "149" and said nothing.

### 3. The "platform sensitivity" premise was false

The fixture said Chromium replay is *"platform-sensitive even at the same pinned browser revision"*, and `linux` excluded four channels on that basis. Measurement falsified it. Same machine, same OS, same capture:

| replay browser | pct |
|---|---|
| 151 (stray) | `99.975379%` — **exactly the old darwin row** |
| 149 (pinned) | `99.558278%` + `flex-line-count`, `position-anchor`, `ruby-overhang`, `text-fit` |

`darwin` was recorded on the stray and needed no exclusions; `linux` ran the pin, saw the four, and excluded them under a plausible-sounding wrong reason. **That false reason is why it survived** — a wrong exclusion that sounds principled is harder to find than a wrong number.

With both platforms on the pin and the four excluded, **darwin and linux are now identical to every digit**: `99.97527632344386`, `4,125,600` cells. There is no measured platform difference left.

### 4. A zero-cell comparison is NULL, not 100

`gate.ts` returned `100` when `cellsCompared === 0` — the exact inverse of the pixel side's own `gate.null-not-zero-means` rule. Naming it in a ledger was not enough: the fabricated 100 **sat in the corpus mean and was counted at ≥90%**.

Consequences, all named rather than absorbed:
- docs mean **86.8% → 86.7%**, **≥90% 59 → 58**, **≥80% 91 → 90** — 14 doc claims updated
- report scripts skip zero-cell cards from means/thresholds **by name and print which**; the component count stays **116**, because an unmeasured component is not ≥90% — the threshold narrows, not the denominator
- `drift-check` gains an **UNMEASURED** state, pinned as strictly as a number (`null → number` or `number → null` is drift and fails), and reads a zero denominator in a committed *harness* scorecard as unmeasured rather than trusting the 100 it still carries (a harness receipt may only be rewritten by a real harness run)

---

### Suspect artifacts produced on the stray browser

1500 committed files mention Chromium 151; almost none are wrong. Separating faithful provenance from suspect numbers:

**Suspect — two classes:**
- **37 mixed-browser captures** (`altitude 8, carbon 10, mui 14, tailwind 5`) — `uaControls` measured on 151, base captured on 149. Arrive on `v1-integration-2` via #45; `main` carries **zero**. Now tracked as **`docs/23 §D.36`** + a fast-lane gate. **Not repaired here** — see below.
- **`computed-floor-platform-baseline.json`** — fixed in this PR.

**Not suspect, deliberately untouched:** `captured-truth.json` provenance (170) is faithful — `capture.ts` stamps the live browser correctly, and that honesty is what made this findable. `scorecard.json`/`numbers.json` (168 each) are harness runs where capture and gate happen in one process on one browser. Contracts and generated `.tsx` carry accurate history. `regate.scorecard.json` / `regate-baseline.json` were already re-recorded under the pin on #49.

### docs/23 §D.36 — open, counted, gated

Repairing the 37 means re-running `ua-baseline-backfill` under the pinned resolver **and** re-recording the drift baseline in the **same round** — the backfill changes what the door subtracts and therefore every `cellsCompared` on those rows. Either alone leaves the tree self-contradictory. That round was deliberately not folded in here.

`npm run mixed-browser:check` (fast lane) prints the per-library breakdown every run and **fails above a committed allowance of 37**. Verified against `v1-integration-2`: it names exactly 37 with the right breakdown. It does *not* fail when the class shrinks — it prints that the allowance is stale, because the honest end state is 0 and tightening it is a reviewed act. Raising the allowance is not the fix.

### Guard

`scripts/render-browser-pin-check.ts` (fast lane), browser-free: it plants fake caches and asks the **real** resolver in a subprocess with `HOME` redirected, so the cache-walking code is what is under test. Falsified — reverting `render.ts`, `gate.ts` or `regate.ts` each turns it red.

### Verification

`225/225` on a clean tree at `f91194a3`, `dirty:false`, measured with **no `PLAYWRIGHT_CHROMIUM_PATH` set** — the pinned resolver chose `chromium-1228` by itself on a machine that also carries the higher stray. `computed-floor-gate` passes on darwin at `99.975276%`, the same number linux gets.

Gates: `tsc` · `lint` · `format:check` · `docs:check` · `capability:fresh` · `extract:computed:drift` · `render-browser:check` · `mixed-browser:check` · `ci:lanes` · `maintain` — all exit 0.

The drift baseline is **not** re-recorded here; it is correct as pushed on `v1-integration-2` (#49). The single `tailwind/Spinner` row changes because this PR changes what the gate emits for it.


tpitre:
Consolidated into #53 — https://github.com/southleft/ds-contracts-poc/pull/53

Merged there onto the new main (828d3e26) alongside #51 (which carries #50). All of it is intact: the pinned-revision resolver, the mixed-browser ceiling (still exactly 37 on the merged tree), the un-fabricated zero-cell row, and the corrected published mean (86.7, >=90% 58/116, >=80% 90/116). Two notes: this branch's docs/23 section is renumbered D.37 there because main's D.36 is #49's control-baseline entry, and its copy of extract/computed/regate-baseline.json predated #49's pinned re-record, so #53 applies only the tailwind/Spinner row delta on top of #49's baseline. Closing in favour of the consolidated PR.

PR #54, opened 2026-08-25
THE HARD CORPUS — cases derived from the closed tables, expectations declared before they run

tpitre:
Stop pulling ten component sets off the shelf from yet another library. Nine libraries × Button/Badge/Card/Checkbox is mostly redundant — every root cause we chase was found on a handful of sets — and the library route pays a tax (sandbox installs, version pins, mixed browsers, hand-edited capture artifacts, captures that are not re-derivable from their own inputs) that has nothing to do with the product. **Author the hard cases instead.** A corpus we author is reproducible by construction, which kills the not-re-derivable class outright.

This PR delivers the design, the gate, and the first measured slice. Nothing was fixed: this round measures, the burn-down fixes.

---

## 1. The corpus is DERIVED, not invented

`spec/hard-corpus.json` is a function of the closed tables, and `npm run hard-corpus:check` re-derives every denominator from its own file and refuses when the committed manifest disagrees — so a mechanism cannot leave the matrix silently.

| table | rows |
|---|---|
| `spec/channel-table.json` | 487 properties — 82 CARRIED / 80 LEDGERED / 271 REFUSED / 54 INERT |
| `spec/grammar-coverage.json` | 44 supported constructs, 10 unsupported |
| `packages/core/src/required-facts.ts` | 61 required facts over 20 archetypes |
| `docs/23-known-limitations.md` | 8 named `FC-*` walls |
| `rootCauses` in the manifest | RC1–RC8 — **the one hand-entered denominator**, transcribed because they live in no repo file |

### The coverage matrix — deliberately unflattering

| source | mechanisms | covered | code→canvas | canvas→code | NOT covered |
|---|---:|---:|---:|---:|---:|
| channel-table CARRIED | 82 | 33 | 25 | 12 (both) | 49 |
| channel-table LEDGERED | 80 | 4 | 3 | 1 (both) | 76 |
| channel-table REFUSED | 271 | 14 | 14 | 0 | 257 |
| grammar-coverage unsupported | 10 | 0 | 0 | 0 | 10 |
| required-facts required | 61 | 0 | 0 | 0 | 61 |
| root causes | 8 | 6 | 6 | 0 | 2 |
| named walls (FC-*) | 8 | 0 | 0 | 0 | 8 |
| **all** | **520** | **57** | | | **463** |

The 54 INERT rows are outside the denominator **by argument** — provably no independent visual effect at computed level, so there is no outcome for a case to expect — not by omission. Every one of the 463 uncovered mechanisms is listed by name in `spec/HARD-CORPUS.md`.

## 2. Not a new directory

`conformance/cases/` is **already** a first-party authored corpus: React + CSS per case, no library dependency, a declared expectation per case, mounted through the unmodified `extract/computed` pipeline, network-free by construction (the harness is built from symlinks; no npm install step exists). The 14 hard cases are conformance cases with a `hard-` prefix. The new `spec/` layer binds them to mechanisms and adds the declare-first rule. Two files now declare each case independently and the gate refuses when they disagree.

## 3. The gate: `npm run hard-corpus:check` (fast lane, with its self-test)

- every case declares `CARRIED | LEDGERED | REFUSED`, and anything but CARRIED must name the string the engine has to produce;
- a result may not exist for a case the manifest does not declare;
- **the commit that first declared a case must be a strict ANCESTOR of the commit that first recorded its result** — same commit is refused. Measured by walking `git log` over both files, not asserted.
- plus shape: sorted, unique kebab ids, homes that exist on disk, a mechanism every source table can be asked about, pinned denominators, byte stability.

`--self-test` plants eleven reds and one green: **11/11**.

The branch history IS the discipline — `6ade6265` declares, `37ce521b` records, and the gate resolves every case to exactly those two commits (see the commit table at the bottom).

## 4. The first slice, measured

14 cases, captured through the unmodified pipeline, pinned Chromium, no library, no install, ~5s each — the whole slice re-derives in 70 seconds.

| case | RC | declared | measured |
|---|---|---|---|
| `hard-flex-direction-column-reverse` | RC1 | CARRIED | **WRONG-NAME** |
| `hard-align-items-baseline` | RC1 | CARRIED | **WRONG-NAME** |
| `hard-svg-glyph-stroke` | RC4 | CARRIED | **SILENT-LOSS** |
| `hard-flex-wrap-wrap` | RC1 | REFUSED | PASS |
| `hard-outline-ring-rest-plane` | RC3 | CARRIED | PASS |
| `hard-box-shadow-layered` | RC3 | CARRIED | PASS |
| `hard-gradient-element-fill` | RC3 | CARRIED | PASS |
| `hard-border-dashed` | RC3 | LEDGERED | PASS |
| `hard-pseudo-tick-rotated` | RC4 | CARRIED | PASS |
| `hard-empty-region-min-size` | RC5 | CARRIED | PASS |
| `hard-placeholder-ink-vs-value` | RC7 | CARRIED | PASS |
| `hard-text-part-pinned-size` | RC8 | CARRIED | PASS |
| `hard-text-indent-eviction` | RC8 | LEDGERED | PASS |
| `hard-rem-scale-padding` | — | CARRIED | PASS |

**11 of 14 came out the way the closed table said they would.**

### RC1, reproduced on an authored case — verbatim receipt

```
root.flex-direction — value shape outside mintable kinds
(color/px/number/shadow/gradient) and outside the declared-channel registry —
no schema channel today (sample: `column-reverse`, 1 distinct value(s))
```

…and the same sentence for `root.align-items` at `baseline`. The channel table classes both **CARRIED**, with a named projection (`layoutMode`, `counterAxisAlignItems`) and a named engine symbol. The keyword reaches the browser, reaches `captured-truth.json`, then leaves through a generic no-schema-channel receipt because the one map that turns a captured flex keyword into a `Part.layout` fact does not carry these spellings. The existing `display-flex` / `flex-gap` cases cannot see it, because a default-direction row is the one value the map does have. **Nine lines of CSS and five seconds — no library version, no sandbox, no browser negotiation.**

### RC4, and it is the worse verdict

`hard-svg-glyph-stroke` grades **SILENT-LOSS** — *observed but neither carried nor named anywhere*. The whole `<svg>` collapses into `root.icon` (an asset reference), the `a` part disappears from the anatomy, and `stroke-width: 2px` is in `captured-truth.json` and in no receipt, refusal, LEDGER line or stdout. A stroked check mark is the most common indicator in every checkbox in every design system. SILENT-LOSS is the class the fixture exists for and the one verdict it never waives; the committed frontier moves 0 → 1 and the gate holds it there.

### One red was mine, and it is named as such

`hard-gradient-element-fill` first graded UNMEASURED because I authored `capturedValue` from raw `getComputedStyle` while the capture reader normalises gradient colour stops to `rgba()`. Repaired in its own commit (`c368b49a`) **before any result existed**, expectation untouched, the wrong value quoted verbatim in the case's `why`, and the lesson written into `spec/HARD-CORPUS.md`.

## 5. Designed, NOT built — said plainly rather than half-built

- **The entire canvas→code half.** No case declares `canvasToCode`. The design is specified in `spec/HARD-CORPUS.md`: one page `hard corpus 2026-08-25` on the scratch file, one set per construct (nested auto-layout with mixed hug/fill, absolute children, component properties, many-axis variant sets, variable modes, text styles, detached instances, overlay-shaped sets), each with a `dump.snippet.json` in the shape the grid corpus already uses.
- **Calendar, data table, combobox** — the archetypes the owner explicitly fears.
- **Multi-axis variant products** — *not reachable* from the fixture as it stands: `conformance/build.ts` writes an EMPTY prop space per case, so no case can carry two enum axes. RC2 and half of RC3/RC4 are products over axes.
- **RC2 and RC6** — RC2 needs an axis; RC6 lives in the computed-capture decision ledger, which only exists for a library that has one.
- **The EMITTER halves of RC5 and RC7** — driven by contract constructs (slots, `placeholder-color`) that no capture produces. They need an authored CONTRACT case on the emit lane; `conformance/canvas.ts` already provides the chain.

## 6. Gates

`tsc --noEmit` · `lint` · `format:check` · `docs:check` · `ci:lanes` · `hard-corpus:check` + `--self-test` (11/11) · `channel-table:check` · `grammar-coverage:check` · `required-facts:check` · `conformance` · `maintain` — all exit 0, serially, in the worktree.

conformance moved 93 cases / 90 pass / 3 red → **107 cases / 101 pass / 6 red**. The three pre-existing UNDECLARED-CARRY rows are untouched.

**Full suite:** `228/228` recorded on a clean tree at `4cc5d0a3` (`commit == HEAD`, `dirty: false`). The first attempt at `37ce521b` came back 226/228 — both reds were derived documents (`conformance/EXPECTATIONS.md`, `docs/24-what-works.md`) that my own 14 cases had moved, regenerated in `4cc5d0a3` with the tools that own them. `docs/24 §3.2` holds the whole conformance fixture out of every library average by name, and that fence held: the real-library headline (116 components, mean 86.8%, coverage 11.1%) is byte-unchanged. An authored corpus must not be able to raise the number the repo quotes about other people's code.

**Five commits, and the history *is* the discipline:**

| | |
|---|---|
| `6ade6265` | 14 cases DECLARED — no capture ran, no result exists |
| `c368b49a` | one FIXTURE REPAIR, named, still before any result |
| `37ce521b` | the results, RECORDED |
| `4cc5d0a3` | two derived surfaces the eval suite caught |
| `a16b14da` | 228/228 |

`npm run hard-corpus:check` resolves every case to `declaredAt 6ade6265 → measuredAt 37ce521b` and would refuse if those were the same commit.


tpitre:
Closing as superseded after the V1 PR audit. The current 126-case conformance frontier covers the important hard-case mechanisms in an evolved form. This branch would reintroduce 230 files absent from main and conflicts with the current conformance baselines. Its coverage-manifest idea remains worth revisiting separately.

PR #55, opened 2026-08-26
BURN-DOWN — 3 root causes landed, 5 open with their refutations, and the bar that blocked them corrected

tpitre:
## The bar, first — because it is why two rounds landed nothing

Rounds 1 and 2 produced **fifteen fixes and zero survivors**. The landing rule
required an adversarial CORRECTNESS verifier and an adversarial VISUAL verifier
to both pass — while the same brief forbade re-minting to Figma and forbade
touching the census `verdict.json`. The census **canvas** PNGs therefore *could
not change*, and the visual verifier refused fixes for exactly that. RC8's
blocking finding, verbatim:

> "THE CANVAS HALF, WHERE BOTH ADDRESSED ROWS' GRADED DEFECTS LIVE, WAS NEVER
> RE-RENDERED … the census canvas PNGs are BYTE-IDENTICAL to the pre-fix ones."

That was the instruction's defect, not the fixer's. An unfalsifiable bar is a
broken instrument. The corrected rule, applied per class in this PR:

- **Correctness is the gate**: root cause addressed; a conformance case
  red-before / green-after *proven by the verifier itself*; no loosened
  assertion, no component-name special case, no silent swallow, no unnamed
  regression in another library.
- **A visual objection blocks ONLY** if it is a genuine NEW visual defect the
  change introduces, or an addressed row that still fails **on the code half**.
  It may not block on "canvas PNG unchanged" (mandated) nor on canvas-only
  defects, which are marked *not assessable until re-mint*.
- **A mis-stated row claim is a claim to CORRECT in the receipt**, not a defect
  that sinks a class.

The whole adjudication, per class, with every refusal quoting its blocking
finding, is appended to `parity/receipts/v1/CANVAS-CENSUS.md` under
**"Burn-down — 2026-08-25"** — in a new PRESERVED LOG region that survives the
gate's own regeneration of that file.

## Landed — 3 of 8

| class | root cause | evidence |
|---|---|---|
| **RC1 layout-axis door** | `LAYOUT_CHANNEL_TO_FIELD` was narrower than the schema it writes into and than both emitters; child ORDER was detected and discarded; a boolean presence axis minted a contract the validator refuses, aborting whole rounds | 5 conformance cases, red→green both directions with a full re-capture; drift 19 → 7 findings |
| **RC8 geometry excluded for text parts** | `absClusterParts` blanket-dropped every text-bearing part from geometry admission for "font-metric-dependent widths" — true of a shrink-to-fit box, false of a box the layout PINNED | 3 cases incl. a harness pin and a wall pin green on BOTH engines; corpus-wide the door admits 2 facts and refuses 10 BY NAME quoting the stage arithmetic; two new freshness gates (`extract:computed:geometry:census`, `bundle:fresh`) |
| **RC7 placeholder ink** | `::placeholder{color}` was READ since R7 and had NO CHANNEL to carry to; three emitter paths, one of which drew nothing at all | new `placeholder-ink:check` (36 pins / 7 cases) + `pseudo-placeholder` UNSUPPORTED→CARRIED + a canvas round-trip row; each half re-proven by reverting ONE file |

## Open — 5 of 8, each on a CORRECTNESS finding

**RC2** the plane gate cannot see a partially-flat axis (mui.link ships 7
identical cell pairs, fluent.avatar 6, both with 0 plane facts) · **RC3** the
verifier BUILT the harm: the ring fold's presence guard is domain-complete over
the DEFAULT plane only, and the folded width is written as `literals` which do
not ride into state previews · **RC4** antd.tooltip still mints an unclipped
rectangular tab on the code half, and fluent.checkbox mixed×circular ships NEW
ink the library does not draw · **RC5** the silent swallow is narrowed, not
closed — a designer's edit on any variant but the first disappears with zero
receipt · **RC6** visual passes, correctness does not: 4 of 31 decision rows are
literals applied with no referee and no name, while the summary reads
`0 applied UNVERIFIED`.

## Four claims corrected rather than shipped

1. **`fluent.tab-list` is NOT carried.** RC1 reported it as needing no work.
   Root layout is `{display:flex, align:stretch}` with no `direction` and
   `layoutByProp: null`; the pre-fix receipt is still live; the canvas compiles
   `Orientation=Vertical` as `mode=HORIZONTAL`; the code half draws the tabs
   side by side. It stays an OPEN row, with its one-command remedy measured.
2. **RC7's `:disabled` placeholder plane is carried by no contract** — verified
   myself: zero state-plane `placeholder-color` in the whole corpus, and
   fluent's six `State=Disabled` cells bind the resting `#707070`. Named per
   cell, so a named wall, not a silent loss.
3. **Two malformed-output shapes in the new CSS lowering** are demonstrated and
   unpinned (unreachable in today's emitters).
4. **carbon.textinput's three dropped zero-width border sides are PRE-EXISTING**
   — I re-fused carbon/TextInput offline on `main`'s engine, which carries none
   of these classes, and it drops the same three keys.

## Base — this PR is stacked, and it has to be

The three classes are based on `fix/rejected-sets` (PR #39) → `census/canvas-full`
(PR #38) → `main`. `main` carries a divergent engine line — its
`extract/computed/fuse.ts` has no `layoutByProp` minting at all — so merging
these classes there would import both open PRs and invalidate every red→green
proof they were measured against. This PR therefore targets the stack, not main.

## Regenerated on the merged engine

All nine new conformance cases RE-CAPTURED through Chromium — every artifact
byte-identical, so the three doors do not disturb each other's fixtures. The
census CODE half re-rendered for all 23 touched rows — every PNG byte-identical.
shadcn's token surface rebuilt as the UNION of RC7 and RC8 (the merge had left
the GENESIS batch unbuildable: `Missing variable: imported/input/root/placeholder-color`).
Fusion-geometry census re-recorded (3 rows EXTENSION-REPAIRED), `golden:update`,
plugin engine receipt, `capability:report`, drift baseline re-recorded.

**Not done, and named:** the canvas half was NOT re-minted and no `verdict.json`
was touched. For the nine rows whose graded defect lives on the canvas, the fix
is verified at the emitted-script and code-render level and is not assessable
until a re-mint-and-blind-re-grade wave runs. That wave is the next round's
first job.


tpitre:
Superseded by #62 — `census/stack-to-main`.

This branch's content is merged there on top of `main`, re-derived on the merged engine and measured: **226 of 230 evals** on a clean tree, with the four reds named in `docs/23` §B.38 rather than re-recorded green.

Note for the record: `burn2/landed` was **not** a descendant of `census/canvas-full` — `git merge-base --is-ancestor` returned false, and all 805 real-library reference PNGs were absent from that head. #62 merges #38 explicitly, and the blind re-grade's **117/170** survives intact along with the 471-line PRESERVED LOG.

Closing in favour of #62.

PR #56, opened 2026-08-26
PROOF — one component through the current engine, before and after, on the canvas

tpitre:
## Verdict: the canvas changed. Yes.

**Before:** the four `Dot=Dot` badges on `Census / altitude` were 36×12 boxes with the word **"Badge"** painted across and out of them.
**After:** four plain **8×8 coloured pips, no text** — which is exactly what the real altitude library renders.

Same node, amended in place: Scratch Project `byMp6lt0Ij9b2QbkDGFwBh` → page **Census / altitude** → set `39:6692`.

Pictures and the full write-up: `parity/receipts/v1/proof/altitude-badge/` (`BEFORE.png`, `AFTER.png`, `REFERENCE-default.png`, `REFERENCE-dot.png`, `PROOF.md`).

| variant | before | after | real library |
|---|---|---|---|
| ×4 `Dot=Default` | 44×20, "Badge" | 44×20, "Badge" | pill + "Badge" — unchanged, right |
| ×4 `Dot=Dot` | **36×12, "Badge"** | **8×8, no children** | 8px dot — **fixed** |

No variant regressed and nothing is wrong in a new way.

## Why the canvas had not changed before this

The engine fix (`Part.textOutOfBox`) has been on `main` since 6ba48b8a. The **contract** did not carry the measured field, so the emitter had nothing to act on: `grep -c textOutOfBox examples/altitude/contracts/badge.contract.json` was `0`.

Re-running the documented chain (capture → promote → emit/validate → `figma bundle`) produced it:

```json
"textOutOfBox": { "prop": "dot", "values": ["dot"] }
```

> *text evidence carried: root.textOutOfBox = {"prop":"dot","values":["dot"]} (MEASURED — text-indent lays the first line at or past the content-box end edge on dot = dot, so the browser paints no text in the box and the canvas draws none either)*

The emitted script now carries **zero** text children on those four combos, and ships the reason as a code-only fact.

## Findings, reported not buried

1. **The altitude corpus was hand-patched.** `chip` and `divider` each carry an `FC-ENUM-HOLE` enum member (`Type=Default`, `Variant=Default`) that no seed carries and no capture can see. A clean re-run silently drops two canvas cells. This branch **restores those two contracts to their `main` bytes** so the proof introduces no regression; the fix belongs in the seeds and is not this branch's job.
2. **`badge.anchors.json` lost all 40 provenance anchors** in the fresh promotion, while library-wide source-aliasing went *up* (41 → 47). Badge's `background-color` moved into `tokensByProp` and the anchor writer did not follow.
3. **`ds-contracts extract --computed` mis-resolves its own repo root** when run through the CLI shell (`REPO` = `packages/cli/dist/../..` = `packages/`), so `tokens.minted` and the bundled fonts are looked for under `<repo>/packages/…` and the run dies before a browser starts. `--root` fixes one path, not the other. The repo-native `npm run extract:computed` — what `PROVENANCE.md` documents, and what this proof used — is unaffected.
4. **Doc numbers re-derived** after the re-capture: 719,534 style cells, 86.8% mean, 86.0% cell-weighted. `npm run docs:check` caught it.

5. **All 8 altitude `.figma.js` scripts had lost their `_prefCol` block.** The block that pins variable lookup to the `"Altitude"` collection was gone from every one of them — which is precisely the damage `scripts/reemit-altitude-figma.ts` warns about in its own refusal text: *"Do NOT re-emit this lane with the CLI `figma` verb — it has no variable-collection option and would strip the 'Altitude' `_prefCol` block."* It had been stripped. Confirmed **pre-existing at the branch tip**, not introduced while fixing this branch: stashing every later change and re-running the check at a clean `62f69dd1` reproduces all 8 stale files. `figma:fresh` lives in the **full** sweep, not the fast lane, which is why this read green locally while CI was red. Re-emitted through the correct script; `GENESIS-BATCH.figma.js` (the artifact a designer actually pastes) inlines those 8 and was rebuilt with it — 9 blocks, 901,047 bytes, 647 variables. Left unfound, this would have surfaced as an inexplicable canvas defect several rounds later.

6. **Finding 1 is one layer higher than reported, and it is an ENGINE DEFECT — not an authoring input.** The `FC-ENUM-HOLE` member is hand-written into the **capture output**, `extract/computed/out/altitude/chip/enriched.contract.json`, so the record claimed a provenance no run reproduces and every downstream artifact inherited it self-consistently. The honest re-capture on this branch erased the patch and made it visible. **Measured:** altitude's enums are all defaultless, so `capture.ts` prepends the `unset` pseudo-value and *renders* it; chip's `captured-truth.json` delta for `secondary.unset__default` vs `secondary.squared__default` differs on **8 border-radius channels** (all `4px` on `squared`, all absent on `unset`). The browser sees the second rendering and the capture measures it — the promoter drops it, emitting `type.enum: ["squared"]`. So this does **not** belong in `authored-facts.json`: that sidecar is for facts the computed truth does not contain, and declaring this un-capturable would assert something the capture disproves. The fix is to give the promoter a spelling for a defaultless axis's `unset` plane; the value is already enumerated, rendered and scored, so what is missing is the **name**, not the measurement. Named in `parity/receipts/v1/corpus-reproducible.json → promote.altitude`; `promote-generalization` stays **RED by choice** until the engine change lands.

## Gates (serial, all exit 0)

`typecheck` · `lint` · `format:check` · `docs:check` · `ci:lanes` · `schema:fresh` · `channel-table:check` · `door-register:check` (+ `self-test`) · `grammar-coverage:check` · `required-facts:check` · `conformance` · `figma:fresh` · `generated:fresh` · `capability:fresh` · `census:check --self-test` · `extract:computed:drift` (VERIFY) · `corpus:reproducible:check` (+ `self-test`) · `maintain`

**The full eval suite HAS now been run** (it had not been when this PR was opened): **229/230**. The single red is `promote-generalization`, which is finding 6 above and is left red deliberately. `evals/results.json` is therefore **not** re-recorded on this branch — `docs:check` refuses a committed record of a red run by name (*"the committed run is RED; docs quoting N/N pass would be a false claim"*), and recording 229/230 would take `docs:check` down with it. The branch keeps `main`'s 230/230 record, which is the posture it already had.

Four artifacts the re-capture moved were re-recorded by their own documented remedies, each with a per-row named cause: `docs/24-what-works.md` (`capability:report`), `extract/computed/regate-baseline.json` + 8 `regate.scorecard.json` (`extract:computed:drift -- --write --config …/altitude.json`), `scripts/door-register-check.ts` `PINNED_CENSUS`, and the altitude figma scripts + genesis batch. All 116 baseline rows kept their prior `gapCause` prose verbatim (108 untouched, 8 extended); every altitude row now agrees with its harness scorecard **exactly** (gap `0.000000`), and **the mismatch count is unchanged on all 8 rows** — every cell the re-capture added is an agreeing cell, so no score here was bought by comparing less.


tpitre:
Closing as historical/superseded after the V1 PR audit. The Altitude findings and provenance gaps are preserved in the current corpus reproducibility receipts, while this branch is over 1,100 commits behind and conflicts broadly with the recipe-IR tree. The branch is left intact for archaeology.

PR #60, opened 2026-08-26
THE DOOR REGISTER'S LINE METADATA WAS 400 FALSE CLAIMS AND NOTHING READ IT

tpitre:
`spec/door-register.json` is the artifact this project cites as evidence that every judgement call in the pipeline is written down. Two of its fields were not evidence.

## The defect

- **415 of 429 doors asserted `markerOutsideRule`** — *"the rule lives inside a serialized in-page function (a template literal), so the marker sits on the nearest line outside it."* Reading every one of them against the tree: **26 do**. 389 false claims.
- **`ruleLine` sat a median of 32 lines off** the line its marker annotates (p25 14, p75 75, max 345), and landed on a **blank or comment-only line 176 times**. 14 doors carried no `ruleLine` at all.
- The cause is mechanical, not sloppiness: `ruleLine` records where each rule sat *before* the `@door` markers were inserted. Computed once, never re-derived.
- **`auditRegister()` read NEITHER field.** That is why it survived four months of a green gate: an artifact nothing reads is an artifact nothing checks.
- `spec/DOOR-REGISTER.md` reprints both numbers in its 429-row table and carried the same staleness — 60 wrong `line` cells, 415 `(rule N)` cells copied from the field that was never derived. The doc half of the gate checked only that each door id *appears* and that the stated count matches.

The door COUNT, the premises, the counter-examples and the receipt paths are untouched. This is a line-metadata repair.

## The repair

**`ruleLine` re-derived for all 429 doors** as *the line the marker annotates*: the first line at or after it that is neither blank nor comment-only. For an ordinary marker that is `line + 1`; prose between marker and rule, and other `@door` markers stacked on the same rule, are stepped over.

**`markerOutsideRule` kept on 26 doors**, each hand-verified by opening the literal and reading the rule inside it — 25 in `capture.ts` (22 in the in-page readers `captureJs` / `capturePortalJs` / `SHADOW_HELPERS_JS` / the animation-pin and autofocus-blur probes; 3 in the generated React harness entry module, which is a written-to-disk module rather than an in-page function and now says so) and 1 in `regate.ts` (the token probe). Each records the real line **inside** the literal, not the declaration line.

The field is a string, and its presence is the boolean: a door that carries it says why; a door that does not is an ordinary adjacent marker. There is no third state.

## What now refuses (direction 5 of the gate)

- a `ruleLine` that is not the line its marker annotates, unless the door claims `markerOutsideRule`;
- a `markerOutsideRule` claim whose `ruleLine` is not strictly inside a multi-line template literal — **parsed with the TypeScript scanner**, so a backtick in a string or a comment cannot fake one;
- any `ruleLine` landing on a blank or comment-only line, claimed or not;
- a door with no `ruleLine`, and a register in which **no** door claims the field at all (the failure mode that hid this one);
- a doc table row whose `line` / `(rule N)` cells are not the register's own, a row count that is not the door count.

Nine new self-test reds, all proved to fire (20 total); the committed register is green.

## Ten doors whose prose does not describe the rule their marker sits on

Re-deriving the lines meant reading the rule at all 429 marker sites rather than the line the register remembered. Ten of them do not say what the code there does. **They are named in the doc, not fixed** — a premise is a judgement call, and rewriting one silently inside a line-metadata repair is the same move that produced the stale lines. Each needs a human to decide whether the marker moves or the premise does. None of them changes what any door subtracts.

| door | the mismatch |
|---|---|
| `fuse.static-part-not-in-union` | **the premise is inverted.** It says "a *contract* part with no *union* counterpart has nothing to probe"; the loop runs over UNION parts and `if (!cPart) continue` drops a union part the CONTRACT lacks. The counter-example it states cannot be produced at that line at all. |
| `gate.unset-axis-deletes-default` | the marker is wedged between an `if` and its `else`, so it annotates `else prop.default = v;` — the **admitting** branch. The deletion it premises is the line above the marker. |
| `anatomy.svg-grammar-fence` | premises "the v1 asset grammar is path/g/circle only"; the marked rule is a computed width/height readability refusal. The grammar fence is 107 lines below. |
| `propose.weight-stamp-differs-refused` | annotates `let weightRef: string \| undefined;` — a bare declaration that refuses nothing. Its rule is 17 lines below. |
| `propose.font-weight-medium-is-default` | annotates `const stampedWeight = […]`; nothing there tests `'Medium'`. The marker is also inserted mid-sentence inside a comment. |
| `propose.weight-not-corpus-nameable` | annotates the branch that **adopts** the stamped ref and returns; the corpus is never consulted on that path. |
| `propose.fixed-size-on-autolayout-node` | the rule it premises is **four lines above the marker**. |
| `propose.grid-overlay-edge-inversion-refused` | the premise is lifted verbatim from a sibling branch about an *unmeasured* child; at the marked branch the child is measured, and the site says outright "the edge spelling is not the blocker". |
| `decisions.ids-are-provenance` | the marker is wedged **inside a `/** … */` block comment**, splitting the module header's own sentence, so the first code it annotates is an `import`. |
| `mint.text-style-namespacing` | same shape — inside a JSDoc block, annotating the interface field `styleName?: string;`. |

## Coordination

Based on `origin/main` at `6ba48b8a`. **PR #58 (`spec/lowering-layout`) is still open** and displaces 167 `@door` marker lines. When it lands, this branch's `ruleLine`s need a re-derivation for those doors — which is now mechanical and, more to the point, **the gate will refuse until it happens** instead of letting the numbers rot silently.

## Gates (serial, all exit 0)

`tsc` · `lint` · `format:check` · `docs:check` · `ci:lanes` · `door-register:check` + `--self-test` · `channel-table:check` · `maintain`

No emitter, engine or corpus file changed — only `spec/door-register.json`, `spec/DOOR-REGISTER.md` and `scripts/door-register-check.ts` — so the full eval suite is not implicated.


tpitre:
## Rebased onto `main` (b834083e) — #58's displacement absorbed, and the remedy is now a command

**The gate did its job.** After merging `origin/main`, `door-register:check` refused: **168 stale `ruleLine`s** and **167 stale doc cells** — exactly the re-derivation predicted in the PR body. That is the failure mode this branch exists to make loud, and it is now a one-command fix rather than the manual procedure I had used.

### `npm run door-register:rederive` (new)

```
    17 door(s) moved in core/propose-figma.ts
    46 door(s) moved in extract/computed/anatomy.ts
   105 door(s) moved in extract/computed/fuse.ts
  168 ruleLine(s) moved; |delta| min 1, median 2, max 4
  ✔ spec/door-register.json rewritten (byte-stable, sorted)
  ✔ spec/DOOR-REGISTER.md table: 167 cell(s) rewritten from the register
```

It is the **same derivation function** the gate enforces, run in write mode — so the check is literally *"run the fix and compare"*, and the two cannot drift apart.

Answering the coordination question directly: **yes, the re-derivation is cheap to repeat.** #61 can land first; whoever merges second runs one command. The gate refuses until they do, so it cannot be skipped by accident.

### The one case a command cannot derive

The 26 `markerOutsideRule` doors point *inside* a template literal, where nothing about the marker's position predicts the line. Each now carries **`ruleText`** — the trimmed source of the line it cites:

- the gate checks that quote against the file, so a citation that carries its quote cannot silently point at the wrong line;
- `--rederive` follows the **text** inside that door's own literal, never an offset;
- where the text is missing, gone, or ambiguous it **refuses by name rather than picking**. `capture.slot-splice`'s anchor occurs **twice inside its own literal**, so this is measured, not hypothetical — and it has a self-test.

An *ordinary* door carrying a `ruleText` is also refused: its line is a function of its marker, so a quote there would be a second source of truth free to rot independently.

### Verified independently, not by the checker that wrote it

A separate re-read of all 429 sites confirms: the marker text at `line` is the door's own; the source at `ruleLine` is code, never blank or comment; nothing but blanks and comments sits between marker and rule; every outside door's quote matches the file. **0 defects.** Twelve moved doors eyeballed by hand against the source.

`markerOutsideRule` needed no revision — #58 touched neither `capture.ts` nor `regate.ts`, and all 26 anchors still read identically.

### Self-test: 25 red cases, all firing

Five new, including the two that prove the remedy is real:

- `--rederive` **REPAIRS exactly what the gate refuses** — all 403 ordinary lines displaced, repaired, and asserted **byte-identical** to the committed register;
- `--rederive` **REFUSES an ambiguous anchor instead of picking one**.

### One more thing

The ten doors whose prose does not describe their rule are now tabulated in the **gate-checked row format**, so their marker lines get re-derived with everything else instead of going stale as hand-written prose. Two moved with #58 (`fuse.static-part-not-in-union` 1733→1737, `anatomy.svg-grammar-fence` 1554→1555) and are now correct. They remain **named, not fixed** — that is the owed follow-up round.

### Gates (serial, all exit 0)

`tsc` · `lint` · `format:check` · `docs:check` · `ci:lanes` · `door-register:check` + `--self-test` · `channel-table:check` · `maintain` — plus `lowering:check` and `canvas:usable:check`, the gates #58 and #59 brought, both green on the merged tree.

Still no emitter, engine or corpus file changed: `spec/door-register.json`, `spec/DOOR-REGISTER.md`, `scripts/door-register-check.ts`, and one line in `package.json` for the new script.

PR #61, opened 2026-08-26
THE LAYOUT RULES ARE IMPLEMENTED — no more centred-row default, the axis vocabulary widens, and margin collapses by axis

tpitre:
Branches from **PR #58**, and now merged forward through **#57, #59, #60, #63, #64, #65, #66, #67** — level with `origin/main` at `34299be5`.

The register's first round changed no behaviour on purpose: it named 57 lowering rules, cited every `proposed` one to the exact line its current behaviour lives at, and stopped. **This round implements the layout family.**

## The dialog on one row, cited to a line and closed

`fui-DialogBody` is `display: grid`. `anatomy.ts` refuses a two-dimensional grid **by name** (`grid-two-dimensional`, and correctly — one axis would have to be invented). But `display: grid` is not in the declared-display vocabulary either, so the container reached the emitter carrying **no layout fact at all**, and `layoutSpec`'s terminal ternary handed it `HORIZONTAL`. Three block-level children — title, content, actions — side by side.

A refusal that falls through to an invention is not a refusal. Two widenings close it:

- **`BLOCK_FLOW_CONTAINER` is now one set at one width** (`block | list-item | flow-root | inline | inline-block`). It used to be spelled twice at different widths — `display === 'block'` at the root, four keywords thirty lines below.
- **Blockification now runs for the root too, and for a container carrying no display fact.** Where there is no display fact only the *safe* half is taken (every in-flow child must be block-level), because "no fact" must never be read as evidence for a column.

### Does it actually un-stack? Measured, before and after

The live usability sweep scored `fluent` 0 of 11 and named `fluent.dialog` as still drawing on one row. Both artifacts run against the same mock canvas the compile receipts use:

| | `dialogbody` | Non-Modal cell |
|---|---|---|
| **before** (`dfb6b8d8`) | `HORIZONTAL` **600×30** — title 144 + body 344 + actions 96 side by side | **676×78** |
| **after** | `VERTICAL` **144×106** — height is the sum of the children | **192×182** |

**What this does not fix, stated rather than assumed:** the live receipt's 1-pixel-tall root is a *different* cause — `dialogsurface__backdrop` is 1×1 and `dialogsurface (margin box)` is `frozen`, so the root's auto-layout lays out neither. The mock renders that root 1400 tall, so the mock and the live canvas disagree there and I will not claim it fixed. The row symptom is closed; the hairline root is not addressed here.

**The centred-row default was measured before it was touched, and deliberately kept.** 16 of 132 committed corpus roots reach it, and every one is a fixed-size control holding nought-to-three inline children — a line box, which is what a centred row *is*. Flipping it would have regressed all sixteen. What changed is the **order**: `layoutSpec` now reads the root's own children *before* it answers.

## What flipped

| rule | was | now |
|---|---|---|
| `emit.axis-layoutless-root-centered-row` | proposed | **implemented** |
| `emit.axis-default-horizontal` | proposed | **implemented** |
| `emit.margin-collapse-summed-not-maxed` | proposed | **implemented** |
| `emit.margin-lone-child-to-padding` | proposed | **implemented** |
| `emit.margin-box-absent-on-amend` | proposed | **implemented** |
| `emit.gap-literal-cross-axis-silent` | proposed | **implemented** |
| `fuse.gap-literal-fallback-misspelled` | proposed | **implemented** |
| `schema.wrap-declared-never-detected` | proposed | **implemented**, renamed `schema.wrap-declared-and-detected` — the old id asserted a fact that is no longer true |
| `fuse.order-refused-before-mint` | — | **new**, the refusal that enforces `schema.order-unregistered-channel`'s wall |

On the merged tree: **58 rules · 45 implemented · 10 proposed · 3 walls · 39 still silent · 36 untested**.

## Two premises corrected by measurement

- **Margin collapse is a property of block flow, not of the axis.** A flex column does *not* collapse. `LayoutSpec.blockFlow` is the discriminator, emitted into the artifact.
- **The single-column-grid lowering was written, then deleted.** `anatomy.ts`'s `lowerGridDisplay` already lowers `cols <= 1`; a second one would have *contradicted a named refusal*. The code now defers and says so.

## RC1, at the cause

`enrichLayout` is the **only** producer of a `Part.layout` axis, and it declined every non-flex container with a bare `continue` — no receipt — reading the **base combo's** display alone while the uniformity check twenty lines below re-read every combo. The display is now read per enabled combo and every decline is named. The vocabulary gains two values **the schema already spells and the emitter already lowers**: `align-items: baseline` and `flex-wrap: wrap` — the latter declared in v15 with a working emitter lowering, a working inverse, and no producer for a year.

## Conformance cases, red before and green after

| case | before | after |
|---|---|---|
| `align-items-baseline` | `WRONG-NAME` — declared CARRIED but NOT carried | **PASS**, and ROUND-TRIPPED |
| `flex-wrap-wrap` | `WRONG-NAME` — declared CARRIED but NOT carried | **PASS**, and ROUND-TRIPPED |
| `layout-on-non-flex-container` | `WRONG-NAME` — named, but not by `layout-container-not-flex` | **PASS** |
| `margin-top-in-flow` (#66's) | **SILENT** — never waivable | **NAMED** |

## A never-waivable red, closed — and it was this branch's rule

#66 asked whether the margin→gap reframe catches `margin-top-in-flow`: a 12px `margin-top` on the second child of a gapless flex column. **It does, exactly** — `lowerMarginGaps` lowers it to a **bound** `itemSpacing` carrying the same token, and the inverse re-emits `gap` as it should. The round trip still read SILENT: not because anything was lost, but because nothing said the margin had become a gap.

The cause was an **ordering bug**, not a missing decision. The pass ran *after* `stripMisses` had emptied the collector, so a `miss()` raised there was never collected — and would have been serialized into every emitted script as a stray `channelMiss` field. Lowering first, collecting second. (Verified: 0 leaked into the 62 re-emitted artifacts.)

`emit.margin-uniform-sibling-to-gap`: receipt `none` → `channelMiss`, and its `roundTrip` verdict **re-derives down** from `round-tripped` to `named`, because the new case is harder than `flex-gap`. A truthful downgrade, recorded as one.

## The five bugs the register recorded — all five closed

`marginBoxCall` missing on the standalone amend path (18 artifacts gained the call) · the summed collapse · the literal cross-axis gap dropped by an `if` with no `else` while the token twin three lines away called `miss()` · the literal-fallback sets spelling the shorthand `gap`, which computed style never reports · `order` minting into a channel no registry knows — the `tab-size` shape that took 32 components down at `validateContract`.

## Merging forward: re-derived, never resolved

Every conflict was a **derived** artifact. `spec/lowering.json` was the only semantic merge, and the two sides were disjoint — no rule touched by both. Three sites came back **ambiguous** and were pinned by hand with the reason recorded. `spec/LOWERING.md` is generated, and the generator was first verified to reproduce **main's** committed doc byte-for-byte from **main's** register (0 diffs over 801 lines). #67's repaired `door-register:rederive` then confirmed the 136 marker lines I had derived by hand: **zero delta on both fields**.

**#66 and this branch repaired the same reader independently.** #66's `eatFlex` is the one kept — it covers everything mine did and `MIRRORED_CHANNELS` is now exported so `channel-table-check` cannot drift from the reader. Two mirrors would be exactly the drift that export prevents, so mine is deleted rather than merged.

## Refused by name, not done

- **`margin: auto`** stays a **wall**. Chromium resolves it to a *used* px value; "centre me" and "push me right" are not recoverable from that number.
- **Asymmetric sibling margins** stay on the `(margin box)` wrapper, registered as the fixed-point violation it is.
- **RC8 (text-part geometry)** untouched, still `proposed` — proving the swap needs a corpus re-capture and the kit has no overlay-plus-text case to red/green it.
- **`stroke` / `stroke-width`** (#66's finding): both are `canvas: "draw"` with no `case` in this file's switch, so the default arm's `if (reg && reg.canvas !== 'draw') miss(…)` is false and the value is neither drawn nor named. **I own that file in this PR and left it alone on purpose** — a fix shipped inside a layout PR is an unexercised claim. I read it the same way #66 does.

## Gates, serial, all exit 0

`tsc` · `lint` · `format:check` · `docs:check` · `ci:lanes` · `lowering:check` + `--self-test` · `channel-table:check` · `door-register:check` + `--self-test` · `canvas:usable:check` + `--self-test` · `required-facts:check` · `conformance` (115 cases, 111 pass / 4 recorded red — all four main's own) · `conformance:roundtrip` (75 cases, **0 red, 0 SILENT**) · `conformance:canvas` · `figma:fresh` (9/9) · `generated:fresh` (3/3) · `ua-baseline:check` · `extract:computed:drift` VERIFY · `maintain` · `eval` **230/230 at `9a641691`, `dirty: false`**.


tpitre:
Closing as superseded. The branch is 939 commits behind main and conflicting; its payload (the layout family of the lowering register) was re-applied on current main by Codex in a sibling worktree and is now committed as `codex/v1-layout-lowering` (14627c325, WIP, unevaluated). That branch is the successor. Both refs are kept for reference. The layout family is not on the v1 path (docs/33 cut the legacy capture engine's spine work); it is revisited after v1.

https://claude.ai/code/session_01ThvpJVww1oHnAo7c7Fo3sq

PR #92, opened 2026-09-14
fix(v1): match coverage numerators to known-size populations

tpitre:
Final V1 honesty audit found a real population mismatch: Fluent contributed 11 covered components to a 113/1015 fraction despite having no measured library-size denominator. This change reports 102/1015 (10.0%) for the eight known-size libraries, keeps the 11 unknown-size covered components in the inventory, and refuses an all-library coverage percentage. Recorded size units remain explicitly heterogeneous, not a harmonized component-family metric.

Also corrects stale current corpus counts (133 scorecards / 11 libraries), golden-file counts (292), the generated report’s hardcoded Six, and the README’s unsupported two-minute full-eval promise. The mean explicitly uses 132 nonempty comparisons; the zero-cell scorecard is not a perfect match. No capture, scorecard, component source, Figma node, owner grade/signoff, or frozen lineage changes.

Proof: the new report regression failed before with 113 != 102, then passed. Four adversarial tests cover unknown-size exclusion, malformed inputs, stale worded/numeric prose, and tampered total/subtotal tables, wired into docs:check. Generator and checker derive the cohort independently. Existing capability eval was observed red on the old caveat wording, then passed with the new repeated cohort/exclusion caveat; no full-suite record was overwritten by that subset run.

Verified locally: docs:check (9 tests plus all doc claims), capability:fresh, targeted capability eval, 25 V1-definition tests, definition check, lane coverage, scoped lint, package formatting, clean diff. Full exact-head CI must pass before merge. Compatibility: reporting/checking correction only, no public contract/API/schema change. No release action or owner approval is fabricated.


tpitre:
Follow-up commit 4e02188fe archives the full lane’s actual eval and readiness records under an exact-SHA artifact name, using the upload action pin already present in this repo. Only four explicit receipt paths are included; no environment files. Archives preserve red receipts too, so consumers must verify commit, dirty flag, and verdict rather than treating artifact existence as success. Missing archive guard reproduced red before the change; explicit-path/hidden-eval/fail-on-missing checks and ci:lanes passed after it. Latest-head lanes are running; no merge until all required checks pass.

PR #113, opened 2026-09-17
Deliver React composition graphs to Figma

tpitre:
## What was broken

The `/sources` workflow could compile a composed React Card graph, but it could not deliver that closed graph through the application to Figma. Nested caller text was treated as an unsupported parent property, dependency components were not created as one operation-scoped graph, and independent readback only understood a single root.

## User outcome and current status

- Outcome and scope: From the React composition review, a user can now prepare, pair, create, and independently inspect a seven-dependency, six-variant native component graph in the authorized Figma evaluation file. Nested Checkbox state forwards to the dependency instance, while caller-owned slot text stays directly editable on the canvas with deterministic contract identity.
- Current milestone and next step: The first React-led composed delivery is structurally complete. The next step is visual and behavior qualification in one cohesive application review, followed by the design-led clean React consumer.
- Public status updated: `README.md` and `docs/CURRENT.md` now distinguish this demonstrated path from general repository onboarding and keep V1's remaining gaps explicit.

## The mechanism

`prepareReactCallerNativePlan` compiles the sealed source graph into an operation-scoped, dependency-first native plan. `createNativeOperationJobs` now admits the exact `react-caller-graph-draft` request and holds one stable reservation for the source composition. The shared writer emits graph target identities; independent readback verifies every dependency main, nested instance property, inherited private descendant, caller slot, and caller text marker. Unsupported representations still fail with `react-caller-native-graph-blocked` before allocation.

## The proof

- [x] A check exists that **fails when this change is reverted**: `source-reference/react-composition.test.ts` exercises caller canvas text lowering and stable graph reservations; `core/composition-scalars.test.ts` exercises graph creation/readback and refusal cases.
- [ ] `npm run ci:lane fast` is green locally — the first run passed 194/197 gates and exposed generated-script whitespace plus shifted lowering anchors; those three gates were fixed and rerun green. GitHub is rerunning the complete lane on the corrected head.
- [ ] `npm run ci:lane full` is green locally, or the failing gates are named below with why — delegated to required GitHub checks.

Additional validation completed:

- `npm run source:reference:check`: 914 tests passed.
- `npm run react:conformance:check`: 98 tests passed.
- `npm run test:playground`: 17 tests passed.
- `npm run docs:check`, root/playground typechecks, lint, plugin check, `npm run figma:fresh`, both lowering gates, Flowbite bundle freshness, and `git diff --check`: passed.
- Live application operation `4e6e22c3-b90d-41e3-bd11-ac50571507ce` created 7 dependency components plus one 6-variant parent set with 169 scoped variables. Independent structure readback passed.
- A Figma canvas instance title was edited independently while its sibling and source main remained unchanged.
- Repeating the application request returned the same completed operation and left exactly one operation page, one parent set, and one proof section.

## Gaps carried forward

Visual fidelity, responsive behavior, interaction qualification, cohesive in-app comparison framing, first-run latency, the design-led clean consumer, and general two-way recovery remain V1 work. This path remains a configured React cohort rather than general repository onboarding.


tpitre:
Continuation review (2026-09-17/18), pushed as three commits on this branch:

- `d4bce0140` — a second read-only readback of the app-created composed Card graph (operation 4e6e22c3, one hour after creation) refused with `native-source-observation-node-inventory`. Figma had re-identified all 66 caller-content nodes inside nested instance slots (`51:2608` → `I51:2604;51:2547;51:2609`); every node was still present with its allocation stamp, type and SLOT parent. The graph verifier now resolves such rows by stamp and slot topology (never names). Nine tamper cases against the real receipt refuse; both saved receipts verify; journal replay flips the stored refusal to `component-structure-observed` without a new write.
- `1d022f5bc` — the application review: unchanged React source beside the native default variant at one image pixel per CSS pixel, all six variants expandable; `caller-react/source-frame` route; export bounds on caller readback; React padding-shorthand warning fixed. Verified in the running app (source current, structure observed, both 360 × 200 px, no console warnings).
- `ddd61e8cd` — from an independent review of this PR: inherited slots the parent leaves unfilled must now mirror the main's slot children (a dropped rectangle is refused; previously any non-born descendant of a born instance was accepted as borrowed). Docs corrected: dependency mains are verified by identity only; caller text in a slot is canvas-editable, never parent-bound; USER-JOURNEYS describes the delivered graph flow; two-level caller content is a named inventory refusal.

Known remaining gaps, stated rather than closed here: the caller delivery path (`prepareReactCallerNativePlan` → `buildReactCallerNativeWrite` → readback) has engine-level mock coverage (`core/composition-scalars.test.ts`) and the live receipts, but no jobs-level test driving `reactCaller.prepare`/`buildComponent` through `createNativeOperationJobs`; the synthetic composition fixture's child text carries a fill token the graph-draft compiler refuses, so that test needs a reshaped fixture. A fresh live readback after the fix still requires the companion plugin to be opened in the Evaluations file. Visual fidelity remains unqualified.

PR #114, opened 2026-09-18
Design-led clean-consumer check: generated React installs, behaves and is scored against Figma; slotless components refuse children

tpitre:
## User-visible outcome

`npm run design:consumer:check -- --dump <rest-dump.json> --contract <proposed contract> --generated <generated dir> --component <Name> --out <evidence dir>` is the first measured consumer step of the design-led journey. It packages `ds-contracts generate` output as an installable library (esbuild transpile, tsc declarations, `exports`, truthful `sideEffects`), `npm pack`s it, installs it with react, react-dom and vite into a temporary consumer with **no path back to this repository**, builds it, mounts every Figma variant in Chromium over loopback HTTP, replaces the TEXT-bound prop and array item text at runtime, switches variants, refuses undeclared `children`, fetches Figma's own PNG of each variant (read-only) and scores the root's layout box against it with the existing whitespace-trimmed pixel scorer at the unchanged 5% limit. Every problem is named in `receipt.json`, which also records the dump's degradations, stub contracts, content sizes and ink coverage.

## Validation (designer-authored Altitude sets, read-only, REST reader)

- **Badge** — installed and built; all 303 token definitions delivered; `text` replaced the rendered text in all five label variants; `variant`/`shape` switching changed every affected cell; five dot variants matched Figma at 0% (8×8 px both sides). **Fails:** five label variants at 15.8–18.8% — same family and weight, but "Badge" advances 41 px in Figma and 36 px in Chromium (content 53×20 vs 57×20).
- **Tabs** (composed, two auto-proposed child stubs) — installed, mounted, rendered three items from the design's `repeat.sample` and replaced their text. **Fails:** the Stretch axis was ledgered inert by the reader, per-side header strokes were dropped by name (`stroke-weights-nonuniform`), the panel's sample content is not carried by a stub; 5.7–6.1% against the limit while drawing 1.4% ink against the design's 12.7%.

Evidence: `recipe/evidence/design-led-consumer/{altitude-badge,altitude-tabs}/`.

## Shared rule (moved to a follow-up)

A component whose JSX never renders `children` accepts them by type and discards them at runtime. Both receipts name this as `children-accepted-but-discarded`. The emitter rule that omits `children` from such components' props types (with test and 31 regenerated components) touches many more byte-pinned artifacts than src/components (storybook generated tree, example libraries, showcase, sync-script freshness, drift-probe counts, eval reds), so it is preserved on `codex/react-refuse-unrendered-children` for its own reviewable PR.

## Remaining limits

One component set per run; composition through stubs carries no sample panel content; declared behavior beyond text/array/variant props and accessibility are not measured; per-side strokes have no contract representation; text-bearing variants stay visually unqualified until the font substrate matches. The in-app review for this path is unfinished. Docs: `docs/USER-JOURNEYS.md` (designer-first "Clean consumer check"), `docs/CURRENT.md` (Figma → reusable React row).


tpitre:
Review fixes pushed (see commit message): unmapped axes / duplicate keys / unexercised axes are named problems; isolation is measured (lockfile + installed dist files grepped for the repo path); FIGMA_TOKEN stripped from child processes; slot detection by anatomy shape; children probe reads every cell; doc numbers corrected (dots 0% at 9×8 vs 8×8, labels 53 vs 57 px with the font available and five text-channel-unsupported degradations, Tabs 5.67%/6.14% at 1.4% vs 12.7% ink). Local fast lane: the only remaining reds are the three input-field sandbox gates, which need the git-ignored recipe/sandboxes node_modules absent from this fresh worktree.

tpitre:
Full sweep on the rebased head failed on byte-pinned artifacts outside src/components (storybook generated tree, astryx/eventz-vars/polaris generated trees, sync-script freshness, drift-probe in-sync counts, three eval rows needing named reds). Those all trace to the children-omission emitter rule, so that rule is split out to `codex/react-refuse-unrendered-children` for a dedicated PR with every pin refreshed and reviewed. This PR now contains the harness, evidence and docs only; both receipts record `children-accepted-but-discarded` as a named failure against the unmodified emitter.

PR #118, opened 2026-09-18
Live two-way updates: an existing native component follows a changed React source; conflicts are named; an unknown write is settled by reading the canvas

tpitre:
## What a user can now complete

Through the local app and the companion plugin, on the **existing** twelve-variant Checkbox set in the Evaluations file (no new allocations at any point):

1. **Change the React source, and the same Figma nodes update.** Editing `disabled:opacity-50` to `-40` used to orphan the native component: a source edit changes the reference id, every operation created from the earlier revision vanished from the app, and the only way forward allocated a second component. Now the app lists such operations as *following another revision* and offers **Follow the current source with the existing …**. After that, **Review compiler update** proposed `0.5 → 0.4` on the same four nodes; the canvas showed 40 %; two independent readbacks agreed. Restoring the source byte-for-byte produced the reverse change on the same nodes.
2. **A conflicting designer edit is refused by name.** 80 % set in Figma on one of those nodes gave preflight `refused`, `native-update-opacity-conflict:14:443`, and no write claim was ever created. (The app previously collapsed every such refusal to one generic string.)
3. **An interrupted write recovers.** The app was SIGKILLed at the instant of write dispatch. After restart the write was *not* re-sent. **Resolve by reading the canvas** found the nodes untouched, returned the update to a fresh preflight, and a second write under its own numbered claim was independently verified.
4. **A repeat run does nothing.** Review, prepare and inspect again: no new plan file, the same operation, one write in the whole journal.

## The rules (all general; nothing component-specific)

- **Source succession** (`source-reference/native-source-succession.ts`): an append-only, hash-chained journal seeded by the creation pin. It records that an existing root or initial-state operation now follows a later *sealed* observation of the same source case. It grants **no write authority**: it only selects which observation the update planner compiles as `desired`. Preflight, guarded apply and independent readback are untouched. Adoption refuses unless that observation is readable from the live unchanged source, and refuses while a written correction is unresolved. Reverting the source is another explicit succession, never a rewrite. A succeeded root anchors the observation it follows; successor observations compile under the existing component contract identity so its variables keep their names.
- **Refuse by name**: the native 409 carries an identifier-shaped reason (never paths or free text); an update refusal keeps the native program conflict identifiers; a changed or removed bound token is named.
- **Unknown write: read, never resend** (`native-update-jobs.ts`): while a write is unresolved the journal admits exactly one other dispatch, a read of the actual nodes with the reader already pinned to that update. *Landed* leads to the usual independent readback; *untouched* (the saved baseline itself) leads to a fresh preflight, and later attempts write `apply-claim.N.json`; anything else is `update-recovery-required`, named. The write result itself still outranks a pending read; one arriving after settlement is journaled as `late-write-result` and can only *raise* recovery when it contradicts the canvas. A companion that died mid-write is handed only the read that settles that same write.
- **The companion resumes a saved connection on open** instead of waiting silently for a click.

## A defect the live run found (fixed here, red to green)

Figma stores opacity as float32: writing `0.4` reads back `0.4000000059604645`. The native program called that a postcondition conflict and its rollback guard then declined to restore. Every earlier correction had used 0.5, 0.25 or 0.125, which float32 holds exactly. Opacity now uses the **exact-or-float32** policy already applied to colours, strokes and grids (planner, host matcher, native pre- and postcondition, rollback guard, readback verifier). The unchanged journal then replays as verified, and a fresh readback agreed.

Also measured rather than assumed: the sandbox ships a *prebuilt* `tailwind.css`. A class edit without rebuilding it renders the disabled Checkbox at full opacity, and the app proposed exactly that (`0.5 → 1`). It was not applied; the complete change (class plus rebuilt stylesheet) was.

## Not demonstrated (recorded in `docs/CURRENT.md` row 3)

- **A design-only change is not carried to the contract or React.** A canvas edit is detected and refused; it is never proposed.
- Updates cover only the bounded correction channels; any other source difference refuses by name.
- Root families seal a content-derived component name at observation time, so an edit to a root family own matrix refuses (`bound-token-change-unsupported`).
- Pre-existing and independent of this PR: re-reading a *parent* operation after a written correction breaks its correction chain (the Button root family has been in that state since a parent readback earlier today).
- Each companion round trip took about a minute (known latency gap).

## Verification

New: `native-source-succession.test.ts` (6), `native-update-write-outcome.test.ts` (7), a float32 case in `core/native-contract-update.test.ts` (fails on the old code). Regression: 149 tests across the update, operation, transport and composition suites. `typecheck`, playground typecheck, `lint`, `docs:check`, `plugin:zip`, `plugin:check`, `plugin:ui-check` and `git diff --check` pass. Engine receipt re-recorded for the core and plugin change. Live evidence is private (`private/two-way-live-2026-09-18/README.md` indexes it); the source sandbox was restored byte-exactly (hashes recorded) and the canvas is back at 50 %.


tpitre:
## Adversarial review of this branch — one Critical found in my own change, fixed in 2a9f6a1e1

I held this PR out of the merge chain and had it reviewed adversarially before it could merge on green. Result:

**Critical (real, and visible in my own live run):** after a canvas read settled a write as *untouched*, the transport advanced by itself to a fresh preflight and a **new write**, while the app text said the write is never sent again. With a second or a stalled poller, both write programs could execute. My live recovery journal shows exactly that automatic advance; it happened to be harmless there (one poller, idempotent opacity), but the rule was broken.

**Fix:** (1) a `begin` handshake: the companion asks the app immediately before executing any write, and is refused once a read has been dispatched to settle that write, so a write that was handed out but held up is provably dead; a write that *had* begun and is not on the canvas yet is never treated as dead. (2) *Untouched* now parks in `update-write-untouched`: nothing is dispatched, retried or preflighted until an operator explicitly re-arms it, and the text says a new write will be sent.

**High:** late write results are now judged by an allow-list, and a contradiction re-enters every chain guard as a written, unverified correction (it previously left `wrote=false`, so the chain ignored it).

**Medium/Low fixed:** one unreadable succession journal no longer takes the whole listing down; positional `instance-N` initial states cannot follow a later observation; adoption requires a settled, observed operation; the companion resumes by itself only to finish something it already holds (the first version removed the consent step); text fill alpha joins the float32 policy.

**Tests:** the reviewer's exact two-poller scenario is pinned at transport level, plus three tests that run the real companion code (lost write, held write marker, refused begin). 169 tests across the affected suites pass; the local fast lane also caught that my new test files were not named by any CI-invoked script (fixed).

**Verified safe by the reviewer:** journal replay including pre-existing journals, `accept()` correlation, the succession design (server-derived pins, seed, A→B→A), `reactIdentity` callers, the `identity` compile key, float32 consistency, and the 409 reason filter (no paths or secrets pass).

The docs row and PR description say *Resolve by reading the canvas … a second write … was independently verified*; with this fix that second write needs an explicit operator action. A second review pass on the fix commit is running; this stays out of the merge chain until it reports.

tpitre:
Second review pass on 2a9f6a1e1: **no Critical or High; conditionally safe**, the condition being that no companion sandbox from before the begin handshake is left running. Fixed in the latest commit so the condition no longer applies: `claim` carries `protocol: 2`, and a journal with the handshake refuses to hand a write to a companion that does not send it (checked before dispatch and before the claim file, so nothing is burned). Also fixed from that pass: resume-on-open is now resume-only and stops when nothing is held; an operation whose succession record is unreadable stays visible under the current source with adoption disabled (hidden, the only path left would have created a second component); late results after an unresolved settlement can raise recovery. Deferred and named in the ledger: a write the companion was allowed to begin and then died before executing stays `update-recovery-required` (stuck, not unsafe) until an operator attestation exists. 171 tests across the affected suites, plugin gates, lint and the lane-coverage gate pass locally. Returning this PR to the merge chain.

tpitre:
Added the fifth proof (design-only), shown live and restored afterwards: a designer set four disabled Checkbox variants to 60 % in Figma; **Read design changes from the canvas** named exactly those four values while the update stayed verified and nothing was written; after the source was changed to render 60 %, the ordinary update path found both surfaces in agreement (`begin` asked, write program `no-op` with nothing changed, readback verified, baseline advanced, zero canvas writes). Limits are stated in the ledger: the tool names the change and verifies agreement, it does not edit hand-written React, and the read exists only on the tip of a correction chain. The live runs also exercised the new `begin` handshake with the real companion. 173 tests across the affected suites pass.

tpitre:
Now also carries the parent re-read guard that was #119 (closed as folded): re-reading a parent operation after a written correction is refused by name, because the creation reader judges the corrected canvas against the creation plan and the correction chain pins the parent journal revision, so that read stranded every later update of the component (it is how the Button root family became unusable on 2026-09-18). Verified live against two corrected operations: refused, no journal event added. Brought up to date with main; the combined head passed the complete local fast lane (197/197).

PR #119, opened 2026-09-18
Refuse re-reading a parent operation once it has a written correction

tpitre:
Stacked on #118 (retarget to main after it merges).

Re-reading a parent operation after a written correction both called the corrected nodes *refused* (the creation reader compares against the creation plan) and moved the journal revision the correction chain pins, stranding every later update of that component. That is how the Button root family became unusable earlier today.

The read is now refused by name, `react-parent-observation-superseded-by-correction`, and the app hides the action and points to the latest correction (**Inspect update again** / **Read design changes from the canvas**). Verified live against two corrected operations: refused, no journal event added.

Not repaired here, and said so in the ledger: the already-affected Button chain. Repairing it needs the chain replayed from the parent journal as it stood when the first correction was planned.


tpitre:
Folded into #118: that branch needed a rebase onto main anyway, so carrying this guard there saves a full CI cycle. The combined head passed the complete local fast lane (197/197) before it was pushed. The change is unchanged: re-reading a parent operation after a written correction is refused by name (`react-parent-observation-superseded-by-correction`), the app points to the latest correction instead, and the already-affected Button chain stays a named gap in the ledger.

PR #121, opened 2026-09-18
A committed visual qualification lane for React → Figma: 44 pairs scored three ways from committed bytes, residuals named, nothing graded

tpitre:
## What this adds

The React → contract → native Figma journey had **no committed visual qualification**: no scorer wiring, no manifest, no ratchet, no npm script — only one-off scripts in a private archive. This adds an offline, deterministic lane, `npm run react:native:fidelity:check`, that re-scores **committed bytes** on every run. It measures; it does not grade. Every artifact says so, and the owner grades.

It reuses the recipe side's existing, owner-adopted instruments **unchanged** (`scoreFidelity`, `FIDELITY_BAR`, the glyph-masked pass, the F1 row policy, a tightening-only ratchet). No tolerance, threshold or existing score is touched.

## Three numbers per variant, none hidden

1. the **historical** ink-trim score exactly as `scoreFidelity` computes it;
2. a **layout-origin-aligned** score (integer translation from recorded origins; fractional translation or a scale mismatch is refused by name);
3. the **glyph-masked** score with the native text boxes as the mask.

| Cohort member | Measured | Pass | Named font residual | Named other | Fail | Historical % | Aligned % | Glyph-masked % |
|---|---:|---:|---:|---:|---:|---|---|---|
| Button initial states | 24 | 17 | 6 | 1 | 0 | 0.722–9.195 | 0.408–2.214 | 0.000 (15 rows; 9 are text-only) |
| Checkbox initial states | 12 | 12 | 0 | 0 | 0 | 0.000 | 0.000 | 0.000 |
| Composed Card, six parent variants | 6 | 6 | 0 | 0 | 0 | 1.344 | 1.210 | 0.000 |
| Retained Card frames | 2 | 2 | 0 | 0 | 0 | 1.344–2.519 | 1.210–2.255 | 0.000 |

The tolerant zeros are **not pixel identity**: exact-pixel difference runs 5.7–99.3 % and the report prints it per row.

## What the seven "failing" Button states actually are

The ledger said "seven Button initial-state exports still exceed the image limit". Measured:

- **Typography is identical on both sides** on all 32 text-bearing pairs (family, weight, size, line height, letter spacing). Advances agree to 0.19 px (Figma hugs to `ceil`). Glyph pixel columns are identical. **No above-tolerance pixel differs outside a native text box.** Chromium lays down 1.12–1.36× Figma's ink for the same glyphs — text antialiasing, which no conversion rule can close.
- **Six** are text-only variants (no background): the ink-trim alignment makes a 1 px fringe the whole denominator. They are named `font-substrate` in the ratchet and satisfy the existing F1 row policy (text-only cell; aligned ≤ 5 %).
- **One** (`disabled=true, variant=secondary`, 6.166 %) is **not** a font residual and is not labelled as one. Its class is `alignment-trim-threshold`, with a proof the check recomputes every run: the native interior composites to rgb(250,250,250) and the source to rgb(249,249,249), one 8-bit step across the shared scorer's white-trim threshold, so the trims diverge (92×14 against 114×34); aligned 0.626 %, glyph-masked 0.000 %. Relabelling that row `font-substrate` turns the gate red by name. The historical 6.166 % stays in the record.

## Not measured, and why (nothing substituted)

**All 63 retained Button root mains.** An original React image exists for every variant, but every native main has an empty content slot while the originals render a label, and neither side records a layout origin (0 of 63 on both sides). No images are committed for them. Making them measurable needs populated instances or recorded origins.

## Evidence and tamper resistance

`recipe/evidence/react-native-fidelity-v1/`: 88 PNGs for 44 pairs, `manifest.json` (sha256 of every byte, operation/update/attempt ids, journal event names), `SCORECARD.json`, `KNOWN-FAILURES.json`, generated `REPORT.md` — 604 KB, no local path, token or file key. `react:native:fidelity:record` rebuilds it from the private journals, verifying each journal's full hash chain, and is byte-identical across runs; it is excluded from CI with that reason. Every committed Button number matches the earlier private measurements 24 of 24, Checkbox 12 of 12.

17 tamper tests, each red **by name**: a changed PNG byte, manifest hash, re-encoded image with a matching new hash, each of the three numbers, a stale report, a moved or shrunk text rect, a moved origin on either side, fractional translation, scale mismatch, ratchet edits (add a passing row, remove a failing one, phantom row, generic class), the trim proof term by term, dropped or re-filed pairs, qualification edits, a moved scorer pin; report bytes are deterministic.

## Stated limits

The six Card variants are paired with an independent nested-Checkbox inspection by property values; that inspection is pinned by its own inventory, not a journal, and the manifest says so. The card-composed frame uses the latest readback. For the two retained frames the font Chromium painted was not recorded. Re-encoding source crops depends on this Node's zlib, so byte-for-byte reproduction on another machine is untested. Interaction states, responsive behaviour and the three caller-content Button comparison frames are not in scope.

## Verification

The complete local fast lane: 198 gates, only the three sandbox-only `recipe:input-field:comparison*` gates (which CI skips) fail. `ci:lanes` accepts the new gate and test wiring. `docs/CURRENT.md` is updated in the three places that described this gap; none claims the criterion is met.


tpitre:
## Added: a root-geometry qualification for the 63 mains that cannot be pixel-scored

The 63 retained root mains stay **not measured for pixels** (empty content slot vs a rendered label; no layout origin on either side; the native update can no longer be re-read with the bounds-capturing reader). Their **non-content box facts** are now compared instead, with no Figma write and no browser work: what the sealed React observation measured on the root against what the native main reads back, joined purely by data (the plan's axes and each main's own `variantProperties`; no component or variant name appears in the logic).

**63 mains × 19 facts = 1,197 verdicts: 1,155 match, 0 mismatch, 42 not comparable.**

- Padding (4 sides), border width (4 sides), radius (4 corners), background paint, border paint, opacity, gap and shadow: 63/63 match each. All 126 native paints are variable-bound, resolve through the readback's own variables and agree with the source on the sRGB 8-bit grid. The nine outline mains carry a five-layer shadow the repo's existing parser accepts.
- **Height:** 56 match (24/32/36/40 px: declared on both sides). The 7 mains with the size prop unset are `not-comparable:content-derived-height-empty-slot` with both numbers printed (22 px in React, 3 px in Figma): the source origin is `auto`, the native sizing is HUG and the plan declares no fixed height.
- **Width:** 28 match (the four icon sizes, declared on both sides) and 35 are `not-comparable:content-derived-width-empty-slot` with both numbers printed.
- 63 of the 80 source rows pair one-to-one; the other 17 omit a prop, which the root-matrix assembler maps to the default, so each collapses onto an explicit-default variant (listed with equal tree hashes).
- No rounding or tolerance was added: the repo's exact-or-float32 policy only, since the writer applies none.

**What a match is not:** the native plan was compiled from these same observations, so this measures *carriage* — that a measured value survived the mint, the variables, the writer and Figma — not an independent second measurement. It proves nothing about width of a content-sized box, text, fonts, antialiasing or any pixel. The report and `docs/CURRENT.md` say so, and nothing here is a grade.

A mismatch must be named in the tightening-only ratchet (currently empty for geometry). 17 more tests (34 total in the two files), each tamper red by name, including flipping a dimension's classification in either direction. The existing `SCORECARD.json` is byte-identical. Complete local fast lane: only the three sandbox-only `recipe:input-field:comparison*` gates fail.

PR #125, opened 2026-09-19
A composed component from an independent family passes through the app: grid roots, declared fill width and a caller frame for fill-width comparisons

tpitre:
Stacked on #123 (it contains that branch; review the last five commits).

## What a user can now complete

A composed component from an independent family went through the application end to end and **passes the visual limit**: shadcn `Alert` (a grid root with a nested title and description), declared by a separate source workspace, no application code naming it. Through the Sources page and the companion in the DS Contracts Evaluations file:

1. four-variant native root family (fill-width root, one-column grid content carrier, 25 variables), structure independently read back;
2. both nested mains (AlertTitle, AlertDescription) created and independently read back; composition ready 2 of 2;
3. composed comparison created and read back: native 360 × 68 against the original 360 × 68, **3.06 % against the unchanged 5 % limit (pass)** with the repository's unchanged scorer; every differing pixel is glyph rendering.

Each step first refused by name on the real input; each refusal became one general rule.

## The rules (no component names)

- **A traced top-level grid root takes the bounded row-flow lowering composed children already had** (one implicit auto column, intrinsic rows, row flow), from sealed grid-constraint evidence gathered per matrix plane. Explicit tracks, areas, dense or column flow, definite rows, RTL, vertical writing, fixed or clamped height refuse by name. Archives saved before this evidence existed re-assemble byte-identically.
- **A root's own `width:100%` is a declared fill**, a new style-origin status given only when the winning own declaration is exactly `100%`, the typed value is 100 percent, the border box equals the containing block's content width, and that containing width is itself definite (a shrink-to-fit place reads `declared-fill-width-containing-block-indefinite`). It projects the parent-width literal the compiler already lowers. A caller `style`/`className` width still refuses (`react-root-grid-width-unqualified`); a measured box is never promoted to a size. No reader of size facts treats `fill` as fixed or auto.
- **A composed child of a grid ROOT takes its width from the root rule**, not from "both sizes auto".
- **A fill-width root is compared inside its caller's observed containing width.** New `containerWidth` (not `instanceWidth`, which pins the instance): it sizes only the app-owned caller frame; the instance stays FILL; the main, its contract and tokens never see the number. Planner, emitted runtime, readback verifier and frame repair agree; two named refusals.

Built-in flows: nested-child derivations, root matrices and fresh plans are digest-identical before and after for the saved built-in archives; the saved card-composed comparison re-prepares identically apart from the random operation id.

## Not claimed

Badge (simple) still misses the limit on glyph rendering alone (6.25 %); Switch (stateful) is created but its independent readback refuses one float32 step on the transparent hit area. The family shares shadcn's stylesheet with the built-in cohort. No update cycle was run on it. Live evidence is private.

## Verification

Tests red before and green after at both the core and source-reference levels (961 in `source:reference:check`), `plugin:check`, `plugin:ui-check`, typecheck (root and playground), lint, `docs:check`, `ci:lanes`; engine receipt re-recorded for the compiler and runtime change; complete local fast lane on the merged tree.


tpitre:
Adversarial review before merge (no Critical or High; three Medium, all fixed in c20fde1f2): a size declaration now counts as the component's own only when the winning selector's subject is carried by the element itself (closes a caller stylesheet rule being recorded as `fill` or `fixed`; all 3,260 saved size facts remain own); a fill width bounded only by the harness stage, body or viewport refuses by name and is never used as a comparison width, the evidence records which ancestor supplied the width, and zoomed, out-of-flow or flex-grown roots no longer record `fill`; grid roots carry the named limitation `grid-tracks-observed-for-this-content-only`. Also: `containerWidth` floor 0.01, the two runtime refusals tested, mixed fill/fixed planes end at `style-prepared`, an unplanned fixed frame is refused. Built-in traces and every saved operation, including the live Alert comparison, re-derive identically. Local lane: 193 of 197; three sandbox-only gates, plus `recipe:live:history:test` which exited non-zero under load with no failing test and passes alone (15,553 of 15,553).

PR #128, opened 2026-09-19
A Figma stroke outside layout is drawn as a ring, not a space-taking border: the designer's padding and token bindings are kept (dump v1.35)

tpitre:
Stacked on #127 (it contains that branch; review the commits from `86d849dc2` on).

## The defect (measured)

A Figma stroke takes no layout space unless the designer turns on `strokesIncludedInLayout`; the emitted CSS `border` always does. On the 72-variant CBDS Badge every outline variant failed on SIZE: hugging variants 4 px too wide, and the 16 px-high small outline rendered 20 px high. Neither reader captured the fact, so the proposer could not lower it, and the census did not even count the loss.

## The decision (owner delegated judgment calls, 2026-09-18; recorded in `docs/23-known-limitations.md` §D.39 with how to reverse it)

Keep the designer's numbers. Padding on the canvas stays the contract's padding, still bound to its spacing token; rewriting it to "padding minus border" was rejected because it destroys the binding. Instead:

- **Readers** (REST and plugin, dump v1.35) capture `strokesIncludedInLayout`; absent in an older dump means not captured, never false.
- **Contract**: a part-level flag under Figma's own name, `strokesIncludedInLayout: false`, only the non-default spelled. `border-width`, `border-color` and per-side widths keep carrying the stroke. Validation refuses the flag without a stroke, or together with per-side colours or a `border-style`.
- **Writer**: has never set the field; every auto-layout frame in pipeline-generated sets reads `true` and no designer-drawn frame in the census does, so absent means "in layout", a generated set proposes back to its own contract, and scripts for every existing contract stay byte-identical (asserted). With the flag it writes the field.
- **Code emitters** (React CSS modules, React inline, web components) draw the stroke as an inset `box-shadow` ring composed before any real shadow, per-side as four layers, token-bound values still `var(--token)`; no `border-style` is emitted for such a part.

## Result, same 5 % limit, same session

CBDS Badge **18 of 72 → 28 of 72**; outline × large 0/12 → 7/12 (median 10.8 % → 3.6 %), outline × small 0/12 → 3/12 (44.5 % → 11.1 %); overall median 9.0 % → 5.5 %, max 51.5 % → 17.1 %. Every outline variant now has Figma's height. What remains is the small-text glyph metric shared with the fill and tonal rows. Altitude Badge has no strokes; Altitude Tabs' designer had turned the flag ON, which is why its border already matched, and its contract is byte-identical.

## Pins moved

Dump grammar 1.34 → 1.35 with the full recipe: embedded plugin dump source, both JSON schemas, door register (three new doors), `lowering.json` citations (rule texts unchanged), engine receipt, census (2,857 → 2,904 carried, 0 silent; this fact was previously uncounted), sync ledger baselines and the six pending markers re-recorded verbatim apart from the grammar clause (verified programmatically; still NOT REVIEWED by the owner, as before).

## Limits (in §D.39)

The HTML preview emitter still draws a space-taking border; a shadow token resolving to `none` would void the composed value in that state; translucent per-side colours double at corners; the plugin's drift snapshot does not list the field.

## Verification

Reader tests on REST and the REAL plugin reader, proposer, each emitter (Chromium-measured; it caught and fixed a default overwriting a carried token), writer round trip; complete local fast lane on the final tree.


tpitre:
Adversarial review before merge: no Critical; two High and five Medium, all fixed or named (commits de74be1c4, 906e60abd, fc3e292df, 8fa05d228). **H1** the inset ring vanished in forced-colors mode (an outline-only control lost its boundary in Windows High Contrast): every ring part now gets an inward `outline` inside `@media (forced-colors: active)` only, no layout space, focus still wins; inline cannot carry a media query and says so. **H2** the new writer line had never run on a real canvas: exercised live on 2026-09-18 in the Evaluations file with a six-component probe (flagged and control), including a `createSlot()` node and a flagged GRID frame: no throw; REST readback: controls read `true` (generated frames are born in layout), flagged roots and the flagged GRID omit the key, flagged roots exactly 4 px narrower and shorter (60×37 → 56×33), the slot 180×60 → 184×64. Also: `border: 0` on every ring part (a nested native element showed its UA border again); unitless `0` composed as `0px`; a shadow token resolving to `none` becomes a no-op layer inside the ring's private variable (the repo's own corpora hold 60+ such tokens) and a token that is `none` in one mode only is refused by name; a consumer `style.boxShadow` on the inline surface now follows the ring instead of deleting it; the token gate's `--_` exemption requires a real declaration; the lowering is registered (`css.stroke-outside-layout-inset-ring`); named limits added (paint order under children, the flag cannot be taken back off by the writer, canvas preview, transitions, print). The reviewer confirmed all 984 tracked contracts emit byte-identical output with the flag absent and generated sets propose back to their own contracts under dump 1.35. CBDS Badge unchanged at 28 of 72. Local lane 196 of 199 (the three sandbox-only gates); figma:fresh and generated:fresh green.

PR #129, opened 2026-09-19
A stateful independent component passes a visual look: descendant own size, translation as end alignment, and re-observable inspections

tpitre:
Stacked on #125 (it contains that branch; review the five commits from `815bf8dda` on).

## Why

The stateful member of the independent family (shadcn `Switch`) was created through the app and **verified structurally, yet looking at the canvas it was wrong** in all nine variants: the thumb was a 1 px-wide hugging frame instead of a 16 × 16 circle, and the checked variants did not move it. The structural readback cannot see either defect. The owner delegated visual adjudication on 2026-09-18; this PR is what that look produced.

## Where the facts were lost

Observation kept both (`width/height: 16px` on the thumb in every state; `translate: calc(100% - 2px)` in the checked states). They were dropped later: the observed-content compiler excludes sample geometry for every part below the root, and style origin was read for component roots only, so nothing proved a descendant's size was its own declaration.

## The rules (no component names)

1. **A part below the root keeps its OWN fixed size**, under the root's discipline: style-origin `fixed`, declared by the element's own selector (the own-selector rule from #125, which treats an ancestor condition such as `.group[data-size=default] .part` as a condition, not the declarer), authored value equal to the used one in every plane where the part exists, no direct text. An outer selector, a percentage, `auto` or an unused declaration leaves it unsized. Evidence is recorded on initial-mount planes only.
2. **A main-axis translation equal to exactly the free space is end alignment for those states**, `start` elsewhere: carried by `layout.justify` plus per-plane layout, no pixel offset minted; free space compared in whole 1/64 px layout units; percentages resolve against the part's own border box. Anything else refuses `react-initial-contract-descendant-translate-unqualified:<reason>` with one of 14 named reasons (not the sole in-flow child, partial free space, cross-axis, rotation or scale, RTL or vertical writing, parent not a single-line flex row, …). Decision and reversal recipe are in the commit message and `source-reference/README.md`. One vocabulary change: per-plane layout may be driven by an optional boolean with an omitted VARIANT plane, the condition token maps already had.
3. **Observer identity.** A complete initial-state inspection records the hashes of the 18 modules that determine what an observation SAYS; it may be observed again only when that identity changed, or when it records none AND today's assembler refuses it for evidence it never gathered. Old runs are kept byte for byte, `latest` moves only on completion, operations pin runs by id so nothing existing changes. This removes a manual step I had to take twice today. The callback-behaviour store is not covered yet (named).

## Live result (Evaluations file, through the app and the companion)

Re-observed under the current observer (9 of 9 combinations), nine-variant set created, structure independently read back in one pass. **Agent visual evaluation of all nine variants against the source: pass** (16 px white circular thumb; left on the light track when unchecked or unset, right on the near-black track when checked; disabled column at 50 %). It is not an owner grade. The earlier defective set is kept in the file as evidence.

## Instrument finding (not fixed; named in `docs/CURRENT.md` row 4)

The repository scorer is blind to a white part on a light track: both fall under its near-white ink threshold, so the six unchecked variants of the DEFECTIVE first set scored 0 % with the thumb missing. Pairing the per-state images with the scorer is also owed: the fidelity recorder is pinned to one reference and the initial readback carries no export bounds.

## Verification

Red before and green after at the assembler level and in real Chromium (the computed forms are `calc(100% - 2px)` and `matrix(1,0,0,1,14,0)`; where the rule says END the part touches the end of the content box; a 10 px translation refuses). Built-in inspections re-assemble byte-identically (six of six) and the saved Button and Checkbox operations re-prepare identically. Whole `source:reference:check` (976), typecheck (root and playground), lint, `docs:check`, `ci:lanes`, `plugin:check`, registers; engine receipt re-recorded; complete local fast lane 195 of 198 (the three sandbox-only gates).


tpitre:
Adversarial review before merge (no Critical; one High, four Medium, all fixed in 30bff5647 and 9c49537f0). The reviewer first confirmed the live correction from the operation's own readback: thumb at x = 1 in the six unchecked or unset variants and x = 15 in the three checked ones, 16 × 16 in all nine, exactly the browser's used offsets; all 984 tracked contracts byte-identical under the vocabulary change. **High:** the own-size rule admitted declared sizes that are not the element's box (the clamp check passes vacuously where `width` does not apply): an inline element, a content-box with padding and border (12 declared, 18 drawn; end to end the END plane landed at x = 19 against the source's 15), `display: contents`, a part under `zoom`. The shared reader (roots and descendants) now requires the MEASURED border box to equal the declared length in whole 1/64 px units and refuses the rest by name (`size-declaration-does-not-apply`, `size-is-content-box`, `size-zoomed-context`); content-box sizes are refused rather than converted, because either number would be wrong somewhere. All 3,472 saved fixed size facts satisfy the new condition. **Medium:** a child's main-axis margins no longer qualify the alignment rule (`main-axis-margin-present`); a descendant moved across states by margins or relative insets refuses `react-initial-contract-descendant-offset-unexplained` instead of compiling silently; boolean-driven by-prop rules were dead selectors on the static HTML surface (exactly one tracked contract changes: a switch whose boolean-driven literals never rendered there); the SUCCESS path of observing again is now tested end to end on a mountable fixture (new run completes, `latest` moves, the old run is byte-identical and an old pin still reads it); a recorded, equal observer no longer strands an unobserved-evidence refusal. Observer identity also records the playwright-core and Chromium version. All eight saved inspections and the live Switch operation re-derive byte-identically. Local lane 195 of 198 (the three sandbox-only gates); figma:fresh and generated:fresh green.

PR #130, opened 2026-09-19
Designer sets that draw only some variant combinations propose: declared absent variants, fail-closed, with an ambiguity fence

tpitre:
Designer-authored component sets with intentionally undrawn variant combinations were rejected as ragged matrices. This change records those missing combinations as complete `bindings.figma.absentVariants` tuples, then requires exact agreement with the declared product minus those tuples. PR #128 has landed; this PR now targets current `main`.

## Behavior and boundaries

- An unstamped designer set may derive absent tuples only when the reader positively observed the shared plugin-data plane. Otherwise the existing ragged-matrix refusal remains, with the additional reason in `SkippedSet.detail`.
- Generated sets use their stamped contract declaration. Missing variants, extra variants and dropped declarations remain named failures. The writer never deletes an existing variant.
- Validation rejects incomplete, duplicate, invalid or noncanonical tuples, absent default combinations, erased axis values, full-product removal and declarations combined with state previews.
- Sparse inference refuses when different minimal axis explanations disagree at an undrawn cell. Mostly-undrawn matrices and products above 4,096 combinations have bounded refusals.
- Native creation omits declared-absent variants. React remains compositional and can render those prop combinations; this limitation is documented.

The consequential choices and reversal instructions are recorded as AGENT decisions in `docs/23-known-limitations.md` §D.40. Frozen held-out evidence is byte-identical to `main`; new diagnostics use the existing detail field without changing historical refusal messages.

## Validation

Exact head `af9da53389960a80af0c98df3c466d9b303c2fdd` completed the full local fast lane: 196/199. The only failures are the three continuation-accepted `recipe:input-field:comparison*` gates. The focused sparse/adversarial suite passed 29/29; historical held-out checks passed with their original bytes. `figma:fresh` (9/9), `generated:fresh` (3/3), type checks and documentation checks passed; the plugin engine receipt was rebuilt.

Read-only designer captures demonstrate sparse proposals for Altitude Checkbox, Progress and Radio. These are projection results, not complete consumer or V1 qualification. Interaction-state interpretation, wider composed consumers and existing visual residuals remain separate work.


tpitre:
Adversarial review before merge: no Critical; two High and three Medium, fixed in 8035f7d23 (tests 17 → 27). What held: the validator is fail-closed; the ambiguity fence gave 0 wrong-axis proposals over an 860-case differential fuzz while the same fuzz with the fence disabled gave 78 wrong proposals in 400; writer output byte-identical on an independent 104-contract sample. What was wrong: (H1) a stamped set whose contract declares an absence while the canvas still draws it read back exact and dropped the declaration; (H2) 'this pipeline wrote it' was decided by whether the reader happened to see a stamp, so a REST response without plugin data laundered a LOST variant into a declaration; (M1) a star set proposed tens of megabytes of tuples; (M3) JSON key order was part of validity. One claim was withdrawn as a consequence of H2: CBDS Alert no longer passes the committed designer exam (that path cannot observe stamps), so the tally stays 5 clean / 19 refused. Investigated and recorded in D.40: Alert's Actions block is present exactly where action is true AND inlineAction is false, a conjunction with a negation that `visibleWhen` cannot express. Local lane on the fixed commit: 196 of 199 (the three sandbox-only gates); figma:fresh, generated:fresh, census and the sync ledger did not move; all 984 contracts identical on six surfaces.

PR #131, opened 2026-09-19
Project designer interaction-state axes with exactness checks

tpitre:
Designer-authored Figma interaction-state axes can now produce exact contract projections when their provenance, axis name and every drawn state are supported. The proposal keeps unknown or ambiguous API values as enum props, refuses missing state behavior, and reports which source cells a later Figma write would add or omit.

The rule uses one closed table across proposal and consumer capture. It requires an unstamped set read through a transport that could observe pipeline stamps, plus an axis named state, states or interaction. Active requires hover or pressed on the same axis. Exact verification derives its expected rows from the contract, checks disabled-name collisions and runs the contract referee on self-contained state-axis proposals. Sparse rest-plane absences remain contract absences; missing non-rest cells are recorded separately. Write-back is not claimed to reproduce the designer's original matrix.

Consumer checks reach hover, pressed and focus with real input, verify that each state was reached and changed supported paint, and release presses without synthesizing a click. This PR also includes the already-integrated consumer measurement correction from #133: whole-pixel capture origins, diagnostic-only text masking, subtree variant-effect observation, transparent node captures and the additional black-background comparison. The 5% limit is unchanged on both backgrounds. Historical captures retain their original scope; the committed Badge remeasurement predates black-background qualification.

Validation after integrating main's merged #130: exact-proposal and browser consumer probes, TypeScript, current design-to-code census, Figma/generated freshness and protected-evidence audit passed. Adversarial checks cover provenance spoofing, missing states, sparse ambiguity, disabled collisions, write-back omissions and ten additional source-axis/row-order permutations. The four frozen CBDS Alert receipts remain byte-identical to main. The complete local fast lane on 51e7147c1 passed 196/199 gates; only the three documented local recipe comparison failures remain. Fresh GitHub CI is required before merge.

Remaining limits: per-state overrides inside nested instances, some unfocusable roots, dead native-disabled selectors on non-native elements, and unsupported stub overrides still refuse or fail their consumer checks. This PR does not qualify a complete family or V1 release. D.41 records the decisions, evidence and reversal instructions.


tpitre:
**HOLD: do not merge.** An adversarial review found one Critical and three High against this PR's own claims; fixes are in progress on the branch. **Critical:** `verified-exact` does not check the state planes against the CONTRACT: the state half of the expected matrix is rebuilt from facts read from the source itself, so it always agrees. On the real sets 29 of 126 rows (23 %) are counted as reproduced while the contract drops them (CBDS Checkbox-icon: 18 of 42, its hover and disabled states; CBDS Toggle 8 of 16); round trip: designer drew 7, pipeline writes 4, both 'verified-exact'. **High:** exact mode returns verified-exact for a contract the validator and emitter refuse (two `disabled` props when a Disabled variant axis sits beside the State axis); `active` maps to `:active` with no guard, while in this repo's own committed CBDS dumps 8 of the 12 axes carrying `active` use it for editing/open rather than pressed (a Tab's selected look would become a mouse-held flash labelled exact); undrawn state cells live only on the proposal, so write-back draws cells the designer never drew (2 for Altitude Menu Item). What held under every probe: the provenance gate, the six named refusals, and the sparse fence; red-before tests are real; 5,279 committed fixtures unchanged (though vacuously for the new branch, which no committed fixture reaches).

tpitre:
Hold lifted. The Critical and the three High findings are fixed in five commits (HEAD 91182afb7), with tests red before and green after; the description above is rewritten to the narrower claim that is now true: Altitude Menu Item and Chip read `verified-exact`, the Menu closure carries its real child, and CBDS Checkbox-icon, Link and both Toggles REFUSE by name (`state-axis-state-not-carried`) instead of being labelled exact. Write-back divergence is named on every proposal. Local lane 196 of 199.

tpitre:
Note: #133 (design-led check: whole-pixel cell origins + text-masked second number; Altitude Badge 10/10, CBDS Badge 46/72) was squash-merged into this branch by the merge chain at 12:53Z because it was stacked here and CLEAN. It had its own complete local fast lane and green CI. This PR therefore now carries #133's changes to main; its CI is re-running on the combined head.

